Issue #52, opened 2024-11-05
MAINT: post migration clean up


Remi-Gau:
suggestions: 
- [ ] convert zenodo.json to citation.cff (do we plan to keep pushing the release to zeonodo?)
- [ ] contributors: I would be in favor of centralizing the listing of contributors in the spec repo so we could get rid of the all_contributors file
- [x] contributing: I can add a file that at least starts point to the general instructions we have on the bids website
- [ ] code climate config contains some lines about the python validator

thoughts?

effigies:
* CITATION.cff: +1
* allcontributors hasn't been actively updated. Either enable the bot or get rid of it, IMO.
* contributing: https://github.com/bids-standard/bids-validator/blob/main/.github/CONTRIBUTING.md
* We don't use code climate. Let's just get rid of it.

Remi-Gau:
okidoki
will take care of those to get some things of your plate

PR #97, opened 2024-11-09
[MAINT] remove code climate and all contributors

Remi-Gau:
- Closes none but relates to #52


Remi-Gau:
also update all contributors bot access: only bids-matlab now uses in the orga

PR #98, opened 2024-11-09
[DOC] point to BIDS COC on website

Remi-Gau:
Unless there is a reason for the validator to have a different COC, I think we should just rely on the BIDS COC, especially that it will 'soon' be updated so having all repos point to it should be enough.


Remi-Gau:
Otherwise we could also just delete the file and the repo will then default to use the one in [.github](https://github.com/bids-standard/.github/blob/main/CODE_OF_CONDUCT.md) repo of the bids organization (which is the one rendered in the website).

effigies:
> Otherwise we could also just delete the file and the repo will then default to use the one in [.github](https://github.com/bids-standard/.github/blob/main/CODE_OF_CONDUCT.md) repo of the bids organization (which is the one rendered in the website).

That sounds good.

Remi-Gau:
COC removed

Issue #76, opened 2019-09-24
Check for opposing SE EPI PhaseEncodingDirections 


dlevitas:
Hello,

We've stumbled upon an issue with a protocol at our imaging facility where the SE EPI PhaseEncodingDirection values are all "j", rather than the PA being "j" and the AP being "j-". When [validating](https://bids-standard.github.io/bids-validator/) the data, no warning messages arose to indicate this issue. From my understanding, SE EPI are prone to error because they can require the user to manually flip the phase encoding direction 180 degrees. Given this, I was wondering if it's possible for the validator to check the SE EPI json files and produce a warning message when the directions are not flipped?

Thanks,

Dan

PR #112, opened 2024-11-14
feat: Extract image orientation from NIfTI header

effigies:
For #76 and #79, it would be helpful to have access to the image orientation. This PR starts with a function to determine axis codes from the the image affine.

If we add `nifti_header.axis_codes` to the context, we could then write #76 as:

```yaml
NiftiPEDir:
  issue:
    code: NIFTI_PE_DIRECTION_CONSISTENCY
    message: Phase encoding direction is inconsistent with entity
    level: warning
  selectors:
    - type(nifti_header) != "null"
    - intersects([entities.direction], ["AP", "PA", "RL", "LR", "IS", "SI"])
    - sidecar.PhaseEncodingDirection
  checks:
    - |
      entities.direction[2 - length(sidecar.PhaseEncodingDirection)]
      == nifti_header.axis_codes[index(["i", "j", "k"], sidecar.PhaseEncodingDirection[0]]
```

(Would need to check that check... It's probably being too clever, but it is nice to get it in one.)


oesteban:
This has bitten me pretty hard; thanks for the initiative.

effigies:
xref https://github.com/bids-standard/bids-specification/pull/2097

effigies:
This is necessary to fully support BIDS 1.10.1. Post-merge reviews are welcome.

PR #119, opened 2024-11-22
Fix bst call for compiling schema

tstoeter:
Without providing a schema path, bst defaults to another installed schema instead of my modifies one:

```
2024-11-22 15:40:41,436 [    INFO] No schema path specified, defaulting to the bundled schema, `/home/tstoeter/.local/lib/python3.11/site-packages/bidsschematools/data/schema`.
```

Making the schema path explicit fixes this.


effigies:
Yeah, this works. I typically have `bst` installed in editable mode, but it's safest to do it your way.

effigies (approved):

PR #124, opened 2024-11-29
feat: Add --prune option to prevent walking subtrees

### LINKED ISSUE #115: Fatal JavaScript out of memory: Ineffective mark-compacts near heap limit
I have been running bids-validator for a few years on an ever increasing dataset. Now I am getting 

==== JS stack trace =========================================

Security context: 0x274c629257c1 <JSObject>
    1: getFiles ...
and in the remaining output is listed a directory name deep inside the /derivatives directory.

FATAL ERROR: CALL_AND_RETRY_LAST Allocation failed - JavaScript heap out of memory

I thought derivatives/ was not searched? It is an old version 0.26.6 so I tried version 2.0.0 using denon and through a bunary download. They both give a similar error

Fatal JavaScript out of memory: Ineffective mark-compacts near heap limit

Is there a way around this memory error?

Thanks for your help

effigies:
This is an optimization to allow for raw datasets with extensive nested derivatives to be efficiently validated. It does not change the default behavior or performance characteristics.

This proactively prunes the filetree during construction, which can prevent many unnecessary stats and construction of many unnecessary objects. The primary downside is that this prevents recursive validation and including pruned files in the dataset size estimate.

Closes #115.

### DIFF
--- a/src/files/deno.test.ts
+++ b/src/files/deno.test.ts
@@ -12,9 +12,11 @@ await requestReadPermission()
 // Use this file for testing file behavior
 const testUrl = import.meta.url
 const testPath = fromFileUrl(testUrl)
-const testDir = dirname(testPath)
+const testDir = dirname(testPath) // $REPO/src/files
 const testFilename = basename(testPath)
+const repoRoot = dirname(dirname(dirname(testPath)))
 const ignore = new FileIgnoreRules([])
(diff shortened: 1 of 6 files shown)


effigies:
I'm not sure if there's any value in performing a similar optimization for the web interface.

PR #127, opened 2024-12-04
fix: Truncate NIfTI header data to avoid extensions

### LINKED ISSUE #126: Large header extensions can result in failure to parse NIfTI headers
I manually decompressed and loaded the first kB of data from a failing file:

```
> readHeader(buf.buffer)
Uncaught RangeError: Offset is outside the bounds of the DataView
    at DataView.prototype.getInt32 (<anonymous>)
    at Function.getIntAt (file:///home/chris/.cache/deno/npm/registry.npmjs.org/@bids/nifti-reader-js/0.6.9/dist/src/utilities.js:28:21)
    at Function.getExtensionsAt (file:///home/chris/.cache/deno/npm/registry.npmjs.org/@bids/nifti-reader-js/0.6.9/dist/src/utilities.js:53:31)
    at NIFTI1.readHeader (file:///home/chris/.cache/deno/npm/registry.npmjs.org/@bids/nifti-reader-js/0.6.9/dist/src/nifti1.js:335:53)
    at readHeader (file:///home/chris/.cache/deno/npm/registry.npmjs.org/@bids/nifti-reader-js/0.6.9/dist/src/nifti.js:146:16)
    at <anonymous>:1:22
```

The relevant lines are here:

https://github.com/bids-standard/NIFTI-Reader-JS/blob/0e406b03775d5687911f9440dffa65d7b0886137/src/nifti1.ts#L386-L412

Because we pass in 540 bytes to allow for NIfTI2, the header parser attempts to parse extensions, and crashes if the extension extends beyond 540 bytes. If we passed 348 bytes in, we would safely ignore the extensions, but break on NIfTI2 images.

https://github.com/bids-standard/bids-validator/blob/9e8da1437b28bf9a69cce3a63727570ec719a754/src/files/nifti.ts#L32-L37

I think probably the right way to fix this is to check if a file is NIfTI 1 or 2 (https://github.com/bids-standard/NIFTI-Reader-JS/blob/0e406b03775d5687911f9440dffa65d7b0886137/src/nifti.ts#L16-L74) and then load the header with the exact number of bytes.

The other approach I can see is to use a try/catch block and fall back to parsing 348 bytes.

Another approach could be to add a flag in NIFTI-Reader-JS to ignore extensions.

effigies:
A couple other issues were identified while debugging.

Closes #126.


src/files/deno.ts
```diff
@@ -93,14 +93,17 @@ export class BIDSFileDeno implements BIDSFile {
...
   async readBytes(size: number, offset = 0): Promise<Uint8Array> {
     const handle = this.#openHandle()
     const buf = new Uint8Array(size)
     await handle.seek(offset, Deno.SeekMode.Start)
-    await handle.read(buf)
+    const nbytes = await handle.read(buf) ?? 0
     handle.close()
-    return buf
+    return buf.subarray(0, nbytes)
   }
```
effigies, line 107: @nellh This seems like it was probably the originally intended behavior?

effigies:
Generally needs review. Somewhat urgent for resolving spurious validation failures on OpenNeuro.

nellh (approved):

Issue #133, opened 2024-12-20
config wildcard ignores: ignored


bpinsard:
I am trying to ignore a set of errors when running the validator on a datalad without any file fetched.

Here is the config I wrote following https://github.com/bids-standard/bids-validator/blob/main/docs/user_guide/command-line.md#configuration-file 

```
{
  "ignore": [
    {"code": "TSV_COLUMN_MISSING", "location": "/**/*_scans.tsv"},
    {"code": "TSV_ADDITIONAL_COLUMNS_UNDEFINED", "location": "/**/*_scans.tsv"},
    {"code": "SCANS_FILENAME_NOT_MATCH_DATASET"}
  ]
}
``` 

But the errors keep showing up (except the last ignore `SCANS_FILENAME_NOT_MATCH_DATASET` ). 
I tried different type of globbing patterns, but none seems to work.
Are wildcard patterns supported? If so in which format?
Thanks!

effigies:
No, the matches are literals.

bpinsard:
From my limited understanding of the code, it seems that the config ignores are merged with `.bidsignore` which can contains globs similar to a `.gitignore`, so that wouldn't work through that either? IIRC the legacy validator is able to handle globs in `.bidsignore`.
How complex would that be to implement globs?
Thanks!

bpinsard:
Ok, do you agree that it would be a useful feature to support glob?
If I understood the code correctly, it would require this function: https://github.com/bids-standard/bids-validator/blob/da2c6e6ace106708c71124a76f6943679cdab80a/src/issues/datasetIssues.ts#L37-L47
to be able to use globs for comparing `location` with something like:
```
import type { Ignore } from '@ignore'
.
.
.

  get(issue: Partial<Issue>): Issue[] {
    let found: Issue[] = this.issues
    for (const key in issue) {
      const value = issue[key as keyof Issue]
      if (!value) {
        continue
      }
      if (key === 'location'){
        let key_ignore: Ignore = ignore()
        key_ignore.add(value)
        found = found.filter((x) => key_ignore.ignore(x[key as keyof Issue].slice(1, x[key as keyof Issue].length)))
      } else {
        found = found.filter((x) => x[key as keyof Issue] === value)
      }
    }
    return found
  }

```
Or we could choose to use the leading slash to distinguish real patterns from globs?
Or prefix globs with a specific character.

PR #134, opened 2025-01-06
enable wildcards in paths for config files

bpinsard:
*warning: I have zero knowledge in typescript*
Attempts at fixing #133. 

Enables usecases where one want to routinely run the BIDS-validator on datalad repo with only the git part checked-out while ignoring all warnings/errors that requires access to files with a generic efficient config file.


src/validators/filenameIdentify.test.ts
```diff
@@ -70,7 +70,7 @@ Deno.test('test hasMatch', async (t) => {
     const parts = tmpFile.split('/')
```
effigies, line 70: Above:

```ts
import { parse } from '@std/path'
```

```suggestion
    const { dir, base }  = tmpFile.split(SEPARATOR_PATTERN)
    const file = new BIDSFileDeno(dir, `/${base}`, ignore)
```

src/validators/filenameIdentify.test.ts
```diff
@@ -70,7 +70,7 @@ Deno.test('test hasMatch', async (t) => {
...
     const file = new BIDSFileDeno(
       parts.slice(0, parts.length - 1).join('/'),
-      parts[parts.length - 1],
+      '/' + parts[parts.length - 1],
       ignore,
     )
```
effigies, line 75: Then you can get rid of this.

src/validators/filenameIdentify.test.ts
```diff
@@ -1,4 +1,5 @@
 import { assertEquals } from '@std/assert'
+import { SEPARATOR_PATTERN } from '@std/path'
```
effigies, line 2:
```suggestion
import { parse } from '@std/path'
```

src/validators/filenameIdentify.test.ts
```diff
@@ -67,12 +68,8 @@ Deno.test('test hasMatch', async (t) => {
 
   await t.step('No match', async () => {
     const tmpFile = Deno.makeTempFileSync()
-    const parts = tmpFile.split('/')
-    const file = new BIDSFileDeno(
-      parts.slice(0, parts.length - 1).join('/'),
-      parts[parts.length - 1],
-      ignore,
-    )
+    const [ dir, base ]  = tmpFile.split(SEPARATOR_PATTERN)
```
effigies, line 71:
```suggestion
    const { dir, base }  = parse(tmpFile)
```

effigies (commented):
Sorry, I meant this...

effigies:
Even though that worked, I think it's not testing exactly what we intended, as `[ dir, base ] = tmpFile.split()` would get the first two directories instead of the directory and the basename.

bpinsard:
I see, so it modifies the tests logic, but it tests what it wanted to test originally: to error on a file that is not in the spec and not ignored by defaults ignores config, right?

rwblair:
Looks like the wonky ignore typing may have been resolved in v7 of the library.

rwblair:
@bpinsard Thank you for the PR!

PR #139, opened 2025-01-09
feat: Process TSV files as streams and validate only the first 1000 rows by default

### LINKED ISSUE #138: Limit TSV loads to a manageable number of lines
Looking at a dataset:

```
$ wc -l sub-20950/beh/sub-20950_task-*.tsv
     457 sub-20950/beh/sub-20950_task-StopSignal_beh.tsv
  315803 sub-20950/beh/sub-20950_task-antisaccade_beh.tsv
     337 sub-20950/beh/sub-20950_task-antisaccade_events.tsv
     154 sub-20950/beh/sub-20950_task-axcpt_beh.tsv
     241 sub-20950/beh/sub-20950_task-flanker_beh.tsv
  316992 total
```

If `**/*antisaccade_beh.tsv` is `.bidsignored`, validation runs in under a second. If not, it takes several minutes at least (I canceled, so I don't know how long).

I think it's reasonable to type check only the first 1000 lines and assume that the rest are fine. We could make this a configurable parameter exposed to the API and CLI, and allow the current behavior to be accessed by setting the value to `-1`. `0` would load only the header, and a positive number would load that many rows beyond the header.

effigies:
This PR is an optimization. In #138 we found a case with >300k lines in a TSV file. In order to limit the number of lines being inspected, I needed to switch TSV loading to be stream-based instead of slurping the entire file.

This PR does the following:

* Refactors the UTF8 enforcement of BIDSFileDeno.text into a stream transformer in files/stream.ts
* Rewrites loadTSV to process files as a stream.
    * UTF8 validation
    * Rechunk as lines of text, splitting on `\r?\n`
* Rewrites column loading as an array of pre-allocated arrays for efficiency. The ColumnsMap is constructed at the end.
* Adds a `--max-rows` flag to the CLI and a `maxRows` variable to validator options.

Note that this adds a new error condition, where we tolerate empty lines only at the end of files (`<content><LF><EOF>`). In passing, this permits us to report the line number of bad TSV lines.

~~I also do not attempt to add `maxRows` to the TSV cache key, so calling `loadTSV()` successively on the same file and different `maxRows` values will return the result from the first call. This does not seem like a problem in terms of running the validator, but might be surprising to future developers. I can look into that, if desired.~~ I added arguments to the cache key in `filememoizeAsync`.

Closes #138.


src/files/tsv.ts
```diff
@@ -2,44 +2,57 @@
...
 import { ColumnsMap } from '../types/columns.ts'
 import type { BIDSFile } from '../types/filetree.ts'
 import { filememoizeAsync } from '../utils/memoize.ts'
-import type { WithCache } from '../utils/memoize.ts'
+import { createUTF8Stream } from './streams.ts'
 
-const normalizeEOL = (str: string): string => str.replace(/\r\n/g, '\n').replace(/\r/g, '\n')
-// Typescript resolved `row && !/^\s*$/.test(row)` as `string | boolean`
-const isContentfulRow = (row: string): boolean => !!(row && !/^\s*$/.test(row))
+async function _loadTSV(file: BIDSFile, maxRows: number = -1): Promise<ColumnsMap> {
```
effigies, line 11: This file in particular will be easiest to review just by reading the new file, as the contents are almost entirely new.

src/utils/memoize.ts
```diff
@@ -15,18 +15,18 @@ export const memoize = <T>(
...
       subcache = new Map()
       cache.set(file.parent.path, subcache)
     }
     let val = subcache.get(file)
     if (!val) {
-      val = await fn.call(this, file)
+      val = await fn.call(this, file, ...args)
       subcache.set(file, val)
     }
     return val
```
effigies, line 32: It looks like in JS there's no equivalent to tuples, so we would need to key on a string.

```diff
-  const cache = new Map<string, Map<F, T>>()
+  const cache = new Map<string, Map<string, T>>()
   const cached = async function (this: any, file: F, ...args: any[]): Promise<T> {
     let subcache = cache.get(file.parent.path)
     if (!subcache) {
       subcache = new Map()
       cache.set(file.parent.path, subcache)
     }
-    let val = subcache.get(file)
+    const key = `${file.path}:${args.join(',')}`
+    let val = subcache.get(key)
     if (!val) {
       val = await fn.call(this, file, ...args)
-      subcache.set(file, val)
+      subcache.set(key, val)
     }
```

Not sure it's worth it.

rwblair: Can kind of kludge tuples with this:
```
let worstTuple: [string, number];
```

Forces first element to be a string and second to be a number.

effigies: The problem is that we would be depending on object equality, which is just identity. We would have nothing but cache misses, IIUC.

PR #140, opened 2025-01-10
rf: Abstract out UTF8-enforcing stream handling

effigies:
Split out from #139 to see if we can track down the resource leak piece by piece. Using CI since I can't reproduce it locally.


effigies:
This is ready for review.

Relatedly, do we want to replace the `BIDSFileBrowser.text()` as well, to keep behavior consistent across platforms? That can be a separate PR.

PR #147, opened 2025-01-31
Add -W (--allow-write) to deno build instructions

yarikoptic:
Without that run which asks to save results into external file would fail with

	error: Uncaught (in promise) PermissionDenied: Requires write access to "derivatives/bids-validator/deno-bids-validator.txt", specify the required permissions during compilation using `deno compile --allow-write`
		at writeFileSync (ext:deno_fs/30_fs.js:899:3)
		at Object.writeTextFileSync (ext:deno_fs/30_fs.js:962:10)
		at main (https://jsr.io/@bids/validator/2.0.1/src/main.ts:38:12)
		at eventLoopTick (ext:core/01_core.js:175:7)
		at async https://jsr.io/@bids/validator/2.0.1/src/bids-validator.ts:2:16

- There is also `--allow-run` which I do not know if should be enabled or not -- e.g. do you run HED validator as external tool, or anything else?

- Is it possible to identify such needs at compile time?

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,8 +31,8 @@ deno run -ERWN jsr:@bids/validator
 ```
 
 Deno by default sandboxes applications like a web browser.
-`-E`, `-R` and `-N` allow the validator to read environment variables,
-local files, and network locations.
+`-E`, `-R`, '-W', and `-N` allow the validator to read environment variables,
+read/write local files, and read network locations.
 
 ### Configuration file
 
--- a/README.md
+++ b/README.md
@@ -27,7 +27,7 @@ The BIDS Validator can be run with the [Deno][] runtime
 (see [Deno - Installation][] for detailed installation instructions):
 
 ```shell
-deno run -ERN jsr:@bids/validator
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -27,7 +27,7 @@ The BIDS Validator can be run with the [Deno][] runtime
 (see [Deno - Installation][] for detailed installation instructions):
 
 ```shell
-deno run -ERN jsr:@bids/validator
+deno run -ERNW jsr:@bids/validator
```
effigies, line 30: Just because RW is a meaningful sequence:

```suggestion
deno run -ERWN jsr:@bids/validator
```

And note that the meanings of `-E`, `-R` and `-N` are described in the next paragraph. Could you add `-W`?

effigies: Also we could mention that `-A` allows all permissions, if this is getting long.

yarikoptic: done for 1st comment. I would not bother with `-A` as not used here/not directly relevant to this PR ;)

PR #150, opened 2025-02-04
feat(expr): Make intersects() return the intersection when non-empty

effigies:
This updates the `intersects()` function in the schema expression language to:

1) Return the intersection, if non-empty. The order will match the order in the longer list.
2) Return `null` on empty intersections to ensure that `intersects()` remains falsey.

This seems like the simplest resolution for resolving bids-standard/bids-specification#914 along the lines of:

> If `participants.tsv` exists, then the `participant_label` entries are a superset of subject directories and participant labels found in `phenotype/`.

With this change, the rule could be written:

```yaml
SubDirSubset:
  selectors:
    - path == '/participants.tsv'
  checks:
    - |
      allequal(
        sorted(intersects(dataset.subjects.participant_id, dataset.subjects.sub_dirs)),
        sorted(dataset.subjects.sub_dirs)
      )

PhenotypeSubset:
  selectors:
    - path == '/participants.tsv'
    - dataset.subjects.phenotype
  checks:
    - |
      allequal(
        sorted(intersects(dataset.subjects.participant_id, dataset.subjects.phenotype)),
        sorted(dataset.subjects.phenotype)
      )
```

If we require that validators sort the arrays in `dataset.subjects` on load, then we can drop the extra `sorted()` calls here.

Because we've seen datasets with subjects in the thousands, I've updated this function to create a Set() based on the smaller list. There is very likely a small n where this constitutes a performance regression.

### DIFF
--- a/src/schema/expressionLanguage.test.ts
+++ b/src/schema/expressionLanguage.test.ts
@@ -16,16 +16,27 @@ Deno.test('test expression functions', async (t) => {
   })
   await t.step('intersects function', () => {
     const intersects = expressionFunctions.intersects
-    assert(intersects([1, 2, 3], [2, 3, 4]) === true)
+    const equal = expressionFunctions.allequal
+
+    const truthy = (a: any): boolean => !!a
+
+    assert(truthy(intersects([1, 2, 3], [2, 3, 4])))
     assert(intersects([1, 2, 3], [4, 5, 6]) === false)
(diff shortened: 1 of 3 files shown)


effigies:
cc @ericearl

effigies:
xref bids-standard/bids-specification#2044

PR #156, opened 2025-02-09
Updated the hed validation to only load schemas once

VisLab:
@rwblair  In looking at `hed.js` it looks as though the `hedSchemas` object is being reloaded for every file being validated. Since the HED schema version specification is the same for an entire BIDS dataset, it need only be loaded once.

@happy5214 (FYI)

### DIFF
--- a/src/schema/context.ts
+++ b/src/schema/context.ts
@@ -39,6 +39,9 @@ export class BIDSContextDataset implements Dataset {
   schema: Schema
   pseudofileExtensions: Set<string>
 
+  // Opaque object for HED validator
+  hedSchemas: object | undefined | null = undefined
+
   constructor(
     args: Partial<BIDSContextDataset>,
   ) {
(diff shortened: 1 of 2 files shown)


VisLab:
@effigies @rwblair:  The reason that this is failing seems to be that the global variable `hedSchemas` is not being reset after each complete BIDS dataset.  There are two ways to fix this:
1. Reset `hedSchemas` after each dataset (I don't know where this would be done in the BIDS.)
2. Alternatively, we could introduce a second global variable that is the current version and check that the new HED version is the same as the one in the global variables and skip the complete rebuilding of the schema after each file is read.

What do you advise?

effigies:
I think the better solution is probably to add some kind of HED validation object (maybe just `hedSchemas`?) to the dataset context, rather than keeping it as a global variable.

VisLab:
I was uncomfortable with the global variable approach.  This is much better.  Thanks!

Issue #145, opened 2025-01-30
Error 'cannot read property null' after validator update


jens-k:
For a dataset that I created last October and that went through the validator perfectly, I now get an error for almost every file. 

The original validation result using an older validator (its still replicable using the [legacy version of the validator](https://bids-standard.github.io/legacy-validator/)) was 
**"This is a valid BIDS dataset!"**

**The [current validator](https://bids-standard.github.io/bids-validator/) throws these errors:**

![Image](https://github.com/user-attachments/assets/39fb698d-59d2-4d04-a5ee-9bae8df5f117)
...

I uploaded **[a reduced dataset, containing only one subject, here](https://www.dropbox.com/scl/fo/c5hlh4euh3h3lr92sd727/AM-xXdPIubRMIccSKUAaQo0?rlkey=mk5lshoarwksb1tzekh00uzut&dl=0)**, so anyone can replicate the issue.

Anyone able to help?

Thanks!

effigies:
cc @happy5214

happy5214:
I could not reproduce this locally. It appears line 77 of `src/validators/hed.ts` may be getting an invalid value, perhaps a `null`. In hed-validator, `BidsFile.validate()` does return `null` for an invalid schema, but why the schema would be invalid on the web and not locally (when it's bundled with the package) is beyond me.

jens-k:
Hi, are there any news on this issue? We will share this issue with the Openneuro people, maybe together a solution can be found?
If there is anything I can do to help from my side, please let me know.

effigies:
No, the error is happening during HED validation. If that's opaque to @happy5214, I don't think I'm likely to solve it quickly without an example dataset.

Can you create a minimal dataset? Given that `participants.tsv` and `participants.json` generate it, could you see if you can remove all but one subject and still produce the error? And then whittle the subject down to as few files as possible without making the error go away? Once you've done that, it should be straightforward to anonymize (truncate the data files to 0 bytes) to the point that the dataset can be shared publicly.


Happy to talk through any of the steps, if you're having trouble.

jens-k:
Hi effigies, thanks for your reply. I thought this is what I did in the [reduced dataset linked above](https://www.dropbox.com/scl/fo/c5hlh4euh3h3lr92sd727/AM-xXdPIubRMIccSKUAaQo0?rlkey=mk5lshoarwksb1tzekh00uzut&dl=0)?

It only has 1 instead of 128 subjects.

[...]

PR #159, opened 2025-02-10
sty: Address deno lint comments for validators/hed.ts

effigies:
Done after reviewing #156.


happy5214:
With a chance to review this now, `file.validate()` shouldn't even be called if the schema is `null`. The validation will never succeed with a `null` schema set (it just returns `null` itself, the probable cause of #145. That needs to be fixed.

effigies:
@happy5214 Done

PR #160, opened 2025-02-12
Clean up HED module issues

happy5214:
This cleans up various minor issues, and it should hopefully deal with the issues in #145 (though, as I couldn't reproduce the original issue, I couldn't test that).


happy5214:
It appears the tests are expecting two HED errors? I only know of one (a failed schema load due to an invalid specification), so I don't know what the other intended error was (perhaps the caught internal error from the spread syntax?). Please investigate.

effigies:
I strongly suspect that the test checked for 2 issues because we were getting 2. We are generally deferring to you on what is expected for HED. Please feel free to update the expected number of issues in the test.

cc @rwblair in case you remember something different.

happy5214:
I confirmed the second issue is from the null pointer, which is no longer generated. I've decreased the issue counts to 1.

VisLab:
Looks good. Everything we discussed has been addressed.

effigies (approved):

PR #162, opened 2025-02-13
fix: Load subjects

effigies:
Type checker doesn't like it, but this fixes the failure to load from TSV files for now. We don't want to synchronously load TSV files during object construction, but nor do we really want to treat subjects as maybe missing.

The real fix I think will be to only call the function once, so we can get rid of the `null` check.


effigies:
@rwblair After this is reviewed and merged, we'll need another release.

PR #166, opened 2025-02-20
fix: Prevent AJV logs from contaminating output

nellh:
Some datasets on OpenNeuro are having JSON output contaminated by this output from AJV's internal logging:

```
strict mode: missing type "array" for keyword "maxItems" at "#/properties/ScreenResolution/anyOf/0/items" (strictTypes)
strict mode: missing type "array" for keyword "minItems" at "#/properties/ScreenResolution/anyOf/0/items" (strictTypes)
strict mode: missing type "array" for keyword "maxItems" at "#/properties/ScreenSize/anyOf/0/items" (strictTypes)
strict mode: missing type "array" for keyword "minItems" at "#/properties/ScreenSize/anyOf/0/items" (strictTypes)
```

Examples include ds005810 and ds005811.

PR #167, opened 2025-02-20
fix: Cast ArrayBufferLikes to ArrayBuffer in read-only contexts

effigies:
Errors seen in #166 and confirmed in `main`: https://github.com/bids-standard/bids-validator/actions/runs/13376538054/job/37571236055

In all of these contexts, the buffers are read-only, so it is safe to claim additional, unused properties.


effigies:
@nellh or @rwblair Does this look reasonable, or should we do something more correct?

PR #168, opened 2025-02-21
chore: Bump @bids/schema v1.0.3 and other dependencies

effigies:
Holding off on `@std/assert`, since that breaks some equality checks.

Builds on #167 to ensure type checking passes.


effigies:
No code changes necessary, merging without review.

If you would prefer I get reviews for things like this, let me know and I'll wait in the future.

PR #170, opened 2025-02-21
fix: Correct Uint8Array / DataView usage for TypeScript 5.7

nellh:
Deno 2.2 via TypeScript 5.7 made typed arrays generic. Specifying the generic for Uint8Array and DataView resolves failures seen in #166.

@effigies I think this is the better alternative to #167.

### DIFF
--- a/src/schema/fixtures.test.ts
+++ b/src/schema/fixtures.test.ts
@@ -11,7 +11,7 @@ const subjectJson = JSON.stringify({ subOverwrite: 'subject', subValue: 'subject
 const rootJson = JSON.stringify({ rootOverwrite: 'root', rootValue: 'root' })
 
 function readBytes(json: string) {
-  return (size: number) => Promise.resolve(new TextEncoder().encode(json))
+  return (size: number) => Promise.resolve(new TextEncoder().encode(json) as Uint8Array<ArrayBuffer>)
 }
 
 export const rootFileTree = pathsToTree([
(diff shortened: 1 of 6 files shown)


effigies:
Doesn't seem to have done the job. Will this continue to work with older deno versions, or will this set a lower bound?

src/schema/fixtures.test.ts
```diff
@@ -11,7 +11,7 @@ const subjectJson = JSON.stringify({ subOverwrite: 'subject', subValue: 'subject
 const rootJson = JSON.stringify({ rootOverwrite: 'root', rootValue: 'root' })
 
 function readBytes(json: string) {
-  return (size: number) => Promise.resolve(new TextEncoder().encode(json))
+  return (size: number) => Promise.resolve(new TextEncoder().encode(json).slice(0, size))
```
effigies, line 14: We either need to overwrite `file.size` or we can just `slice()` to get the full length.

```suggestion
  return (size: number) => Promise.resolve(new TextEncoder().encode(json).slice())
```

src/schema/fixtures.test.ts
```diff
@@ -34,4 +34,5 @@ const anatFileTree = subjectFileTree.directories[0].directories[0] as FileTree
 
 export const dataFile = anatFileTree.get('sub-01_ses-01_T1w.nii.gz') as BIDSFile
 const anatJSONFile = anatFileTree.get('sub-01_ses-01_T1w.json') as BIDSFile
-anatJSONFile.readBytes = (size: number) => Promise.resolve(new TextEncoder().encode(anatJson))
+anatJSONFile.readBytes = (size: number) =>
+  Promise.resolve(new TextEncoder().encode(anatJson).slice(0, size))
```
effigies, line 38:
```suggestion
anatJSONFile.readBytes = (size: number) =>
  Promise.resolve(new TextEncoder().encode(anatJson).slice())
```

nellh:
> Doesn't seem to have done the job. Will this continue to work with older deno versions, or will this set a lower bound?

It will only set the lower bound for tests or running with --check.

Issue #176, opened 2025-03-05
`opaque` flags are not currently used by the validator


bclenet:
The validator does not currently take into consideration the `opaque` flags, which indicate whether the contents of a directory are specified or not.

For example, the following lines in the BIDS schema (file `bids-specification/src/schema/rules/directories.yaml`) describe a `code` directory which contents are not specified.
```yaml
code:
  name: code
  level: optional
  opaque: true
```

Instead, the validator uses as a workaround : the *opaque* directories are inside the list of default ignores in `src/files/ignore.ts`:
```typescript
const defaultIgnores = [
  '.git**',
  '.*',
  'sourcedata/',
  'code/',
  'stimuli/',
  'log/',
]
```

rwblair:
@bclenet Thank you for the issue.

Notes to myself on the problem:
Currently all opaque directories have fixed names. We could treat these like data "files" whose extensions indicate that they are actually directories as done in [walk.ts](https://github.com/bids-standard/bids-validator/blob/main/src/schema/walk.ts) using psuedofiles and the corresponding check for them in the dataset context.

Directories that don't use `name` need more sophesticated validation and should probably get a contextl and some note in the context that they are a directory. Then filename identify and validate can be extended to check the rules/directories.yaml entries. This would also allow us to generate errors for nonbids directories that have no files in them.

If we ever get an opaque directory with a sophisticated naming scheme, requirement level, or is contextually nested I'm not sure what we'd do to prevent all of its children from being ingested in the walk process.

Also to think about: another location that code and sourcedata are used non schematically:
https://github.com/bids-standard/bids-validator/blob/main/src/validators/bids.ts#L78

PR #180, opened 2025-03-10
feat: Validate directories, do not recurse into schema-declared opaque dirs

rwblair:
#176

### DIFF
--- a/src/schema/walk.ts
+++ b/src/schema/walk.ts
@@ -18,11 +18,11 @@ const nullFile = {
   readBytes: async (size: number, offset?: number) => new Uint8Array(),
 }
 
-function pseudoFile(dir: FileTree): BIDSFile {
+function pseudoFile(dir: FileTree, opaque: boolean): BIDSFile {
   return {
     name: `${dir.name}/`,
     path: `${dir.path}/`,
-    size: [...quickWalk(dir)].reduce((acc, file) => acc + file.size, 0),
+    size: opaque ? [...quickWalk(dir)].reduce((acc, file) => acc + file.size, 0) : 0,
     ignored: dir.ignored,
     parent: dir.parent as FileTree,
     viewed: dir.viewed,
@@ -40,8 +40,15 @@ async function* _walkFileTree(
   }
   for (const dir of fileTree.directories) {
     if (dsContext.isPseudoFile(dir)) {
-      yield new BIDSContext(pseudoFile(dir), dsContext)
+      yield new BIDSContext(pseudoFile(dir, true), dsContext)
+    } else if (dsContext.isOpaqueDirectory(dir)) {
+      let context = new BIDSContext(pseudoFile(dir, true), dsContext)
+      context.directory = true
+      yield context
     } else {
+      let context = new BIDSContext(pseudoFile(dir, false), dsContext)
+      context.directory = true
+      yield context
       yield* _walkFileTree(dir, dsContext)
     }
   }
--- a/src/schema/walk.ts
+++ b/src/schema/walk.ts
@@ -39,16 +39,12 @@ async function* _walkFileTree(
     yield new BIDSContext(file, dsContext)
   }
   for (const dir of fileTree.directories) {
-    if (dsContext.isPseudoFile(dir)) {
-      yield new BIDSContext(pseudoFile(dir, true), dsContext)
-    } else if (dsContext.isOpaqueDirectory(dir)) {
-      let context = new BIDSContext(pseudoFile(dir, true), dsContext)
-      context.directory = true
-      yield context
-    } else {
-      let context = new BIDSContext(pseudoFile(dir, false), dsContext)
-      context.directory = true
-      yield context
+    const pseudo = dsContext.isPseudoFile(dir)
+    const opaque = pseudo || dsContext.isOpaqueDirectory(dir)
+    const context = new BIDSContext(pseudoFile(dir, opaque), dsContext)
+    context.directory = !pseudo
+    yield context
+    if (!opaque) {
       yield* _walkFileTree(dir, dsContext)
     }
   }
--- a/src/validators/filenameIdentify.ts
+++ b/src/validators/filenameIdentify.ts
@@ -12,7 +12,7 @@
  * object in the schema for reference.
  */
 // @ts-nocheck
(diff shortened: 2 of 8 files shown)


src/schema/walk.ts
```diff
@@ -40,8 +40,15 @@ async function* _walkFileTree(
...
+    } else if (dsContext.isOpaqueDirectory(dir)) {
+      let context = new BIDSContext(pseudoFile(dir, true), dsContext)
+      context.directory = true
+      yield context
     } else {
+      let context = new BIDSContext(pseudoFile(dir, false), dsContext)
+      context.directory = true
+      yield context
       yield* _walkFileTree(dir, dsContext)
     }
```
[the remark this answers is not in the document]
rwblair: Nice, much cleaner.

rwblair:
Previous commit made me realize we don't generate an error for non `/` path sperators.

effigies (commented):
I've reviewed the walking code and it makes sense. Still need to wrap my head around findDirRuleMatches.

src/validators/filenameIdentify.ts
```diff
@@ -26,12 +26,56 @@ const CHECKS: CheckFunction[] = [
...
+  Object.keys(schema[schemaPath]).map((key) => {
+    const path = `${schemaPath}.${key}`
+    const node = schema[path]
+    if ('name' in node) {
+      if (node.name === context.file.name.replaceAll('/', "")) {
+        context.filenameRules.push(path)
+      }
+    }
+    if ('entity' in node) {
+      let entityDef = schema[`objects.entities.${node.entity}`]
```
effigies, line 59: I would suggest dereferencing `objects.entities` (as well as `objects` for `node.value`) outside the loop.

src/validators/filenameIdentify.ts
```diff
@@ -26,12 +26,56 @@ const CHECKS: CheckFunction[] = [
...
+      // kludge, entries in schema.objects are plural, value specified as singular
+      // will fail for modalities
+      Object.keys(schema[`objects.${node.value}s`]).map((key) => {
+        const value = schema[`objects.${node.value}s.${key}`].value
+        if (`${value}` === context.file.name.replaceAll('/', "")) {
+          context.filenameRules.push(path)
+        }
+      })
+    }
+  })
```
effigies, line 76:
```suggestion
  }
```

src/validators/filenameIdentify.ts
```diff
@@ -26,12 +26,56 @@ const CHECKS: CheckFunction[] = [
...
+    }
+    if ('value' in node) {
+      // kludge, entries in schema.objects are plural, value specified as singular
+      // will fail for modalities
+      Object.keys(schema[`objects.${node.value}s`]).map((key) => {
+        const value = schema[`objects.${node.value}s.${key}`].value
+        if (`${value}` === context.file.name.replaceAll('/', "")) {
+          context.filenameRules.push(path)
+        }
+      })
```
effigies, line 74: Again, I think we want to be able to break instead of iterating over everything

```suggestion
      for (const valueObj of Object.keys(schema[`objects.${node.value}s`])) {
        if (valueObj.value === context.file.name.replaceAll('/', "")) {
          context.filenameRules.push(path)
          break
        }
      }
```

src/validators/filenameIdentify.ts
```diff
@@ -26,12 +26,56 @@ const CHECKS: CheckFunction[] = [
...
+  let schemaPath = 'rules.directories.raw'
+  if (context.dataset.dataset_description.DatasetType === 'derivative') {
+    schemaPath = 'rules.directories.derivative'
+  }
+  Object.keys(schema[schemaPath]).map((key) => {
+    const path = `${schemaPath}.${key}`
+    const node = schema[path]
+    if ('name' in node) {
+      if (node.name === context.file.name.replaceAll('/', "")) {
+        context.filenameRules.push(path)
```
effigies, line 55:
```suggestion
        context.filenameRules.push(path)
        break
```

src/validators/filenameIdentify.ts
```diff
@@ -26,12 +26,56 @@ const CHECKS: CheckFunction[] = [
...
+      }
+    }
+    if ('entity' in node) {
+      let entityDef = schema[`objects.entities.${node.entity}`]
+      if (
+        entityDef && 'name' in entityDef && context.file.name.startsWith(`${entityDef['name']}-`)
+      ) {
+        context.filenameRules.push(path)
```
effigies, line 63:
```suggestion
        context.filenameRules.push(path)
        break
```

effigies (commented):
Part of me thinks we should be able to optimize this further, but in general there will be an order of magnitude fewer directories than files, so I don't want to go too hard unless this turns out to be a noticeable bottleneck.

Some smallish optimizations to short-circuit loops and avoid repeated recursive lookups.

src/validators/filenameIdentify.ts
```diff
@@ -26,12 +26,56 @@ const CHECKS: CheckFunction[] = [
...
   }
 }
 
+export async function findDirRuleMatches(schema, context) {
+  const datasetType = context.dataset.dataset_description?.DatasetType || 'raw'
+  const schemaPath = `rules.directories.${datasetType}`
+  const directoryRule = schema[schemaPath]
+  const schemaObjects = schema['objects']
+  const schemaEntities = schema['objects.entities']
+  for (const key of Object.keys(directoryRule)) {
```
effigies, line 48:
```suggestion
  loop: for (const key of Object.keys(directoryRule)) {
```

src/validators/filenameIdentify.ts
```diff
@@ -26,12 +26,56 @@ const CHECKS: CheckFunction[] = [
...
+        break
+      }
+    }
+    if ('value' in node) {
+      // kludge, entries in schema.objects are plural, value specified as singular
+      // will fail for modalities
+      for (const valueObj of Object.keys(schemaObjects[`${node.value}s`])) {
+        if (valueObj === context.file.name.replaceAll('/', "")) {
+          context.filenameRules.push(path)
+          break
```
effigies, line 72:
```suggestion
          break loop
```

effigies (commented):
Small optimization

effigies (approved):
LGTM. @rwblair merge if you're all set.

PR #181, opened 2025-03-12
Provide tool to build and distribute deno-compiled binary on PyPi

### LINKED ISSUE #148: Upload cross-platform builds of deno bids-validator as .whl's to pypi
Not my idea... I think @effigies and/or @nellh had it: similarly to [uv](https://github.com/astral-sh/uv) which provides .whl files across platforms (see https://pypi.org/project/uv/0.5.26/#files), it would be nice to have bids-validator builds (pretty much output binary from `deno compile -ERNW -o bids-validator jsr:@bids/validator`) to be wrapped into a "full fledged" whl.

Anyone has some other project which does something like that?  I looked at `uv` but it seemed to rely on rust specific https://github.com/PyO3/maturin-action to build whls.

May be @jwodder have heard about anything or has good pointers for easy creation of whls for binaries?

candleindark:
This PR closes #148. Please see commit messages for details.

**Note**:
1. The `--allow-run` option is used to compiled the BIDS validator from local source. Without this option, the compilation result would require the user to grant permission to run subprocesses. This option was not used/needed when compiling from the published jsr package from some reason. 
2. For publishing to work, please set up a github secret `PYPI_TOKEN` containing the PyPi token.

TODOs:
- [x] created https://pypi.org/project/bids-validator-deno/ by uploading 2.0.3 "manual" build done on my @yarikoptic linux laptop . Required "faking" this PR branch to be on 2.0.3 tag for proper tagging. Not sure if we need/want to upload more wheels similarly -- this workflow I think would not work unless we make it force similar nastiness somehow
- [x] invited @effigies @tsalo @candleindark to maintain this PyPI project
- [x] BTW -- I also applied for BIDS community organization on PyPI FWIW  
- [x] add PYPI_TOKEN
- [x] remove mentioning of testing pypi
- [x] validate publishing on test PyPi with the final setup. https://test.pypi.org/project/bids-validator-deno/2.0.12/ 
- ~~set up publishing with OpenID Connect(POSTPONED)~~

### DIFF
--- a/.github/workflows/pypi-whl.yml
+++ b/.github/workflows/pypi-whl.yml
@@ -1,0 +1,73 @@
+name: PyPI Whl Publish
+
+on:
+  push:
+    branches: [main, dev]
+    tags: ['*']
+  pull_request:
+    branches: [main, dev]
+  workflow_dispatch:
+
+defaults:
+  run:
+    shell: bash
+    working-directory: tools/pypi-whl
+
+jobs:
+  build_wheels:
+    name: Build whl on ${{ matrix.os }}
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
+        # windows-latest is an Intel runner
+        # macos-13 is an Intel runner, macos-latest is an Apple Silicon runner
+        # ubuntu-latest is an Intel runner, ubuntu-24.04-arm is an arm runner
+        os:
+          - windows-latest
+          - macos-13
+          - macos-latest
+          - ubuntu-latest
+          - ubuntu-24.04-arm
+      fail-fast: false
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+      - uses: denoland/setup-deno@v2
+        with:
+          deno-version: v2.x
+      - name: Set up Python
+        uses: actions/setup-python@v5
+        with:
+          python-version: 3.9
+      - name: Install build and twine
+        run: python -m pip install build twine
+      - name: Compile deno binaries
+        run: deno compile -ERNW --allow-run -o dist/bids-validator ../../src/bids-validator.ts
+      - name: Print out the content of the dist folder after building validator binary
+        run: ls -l dist/
+      - uses: actions/upload-artifact@v4
+        with:
+          name: bids-validator-${{ matrix.os }}
+          path: |
+            tools/pypi-whl/dist/bids-validator
+            tools/pypi-whl/dist/bids-validator.*
+      - name: Build wheel
+        run: python -m build --wheel
+      - name: Print out the content of the dist folder after building the wheel
+        run: ls -l dist/
+      - name: Validate the built wheel through installation and running the validator
+        run: ./validate_wheel.sh
+      - name: Check if distribution's description will render correctly on PyPI
+        run: twine check dist/*.whl
+      - uses: actions/upload-artifact@v4
+        with:
+          name: wheel-${{ matrix.os }}
+          path: tools/pypi-whl/dist/*.whl
+      - name: Publish wheels if corresponds to a version tag
+        if: github.ref_type == 'tag'
+        run: ./publish.sh
+        env:
+          TWINE_USERNAME: __token__
+          TWINE_PASSWORD: ${{ secrets.PYPI_TOKEN }}
+          TAG: ${{ github.ref_name }}
--- a/tools/pypi-whl/README.md
+++ b/tools/pypi-whl/README.md
@@ -1,0 +1,33 @@
+# bids-validator-deno
+
+**bids-validator-deno** is a Python installer (distributed as wheels on [PyPI](https://pypi.org)) 
+for the self-contained binary of 
+[bids-validator](https://github.com/bids-standard/bids-validator), 
+compiled with [`deno compile`](https://docs.deno.com/runtime/reference/cli/compile).
+-----
+
+## Table of Contents
+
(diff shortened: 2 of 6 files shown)


.github/workflows/pypi-whl.yml
```diff
@@ -0,0 +1,96 @@
...
+          path: tools/pypi-whl/dist/*.whl
+      - name: Publish wheels if corresponds to a tag
+        # TODO: might be suboptimal since we might skip a tag if not the latest commit
+        run: |
+            cd tools/pypi-whl
+            # if a clean tag, no - in it
+            if git describe | grep -e -; then 
+                echo not releasing; 
+            else
+                twine upload --verbose --repository testpypi dist/*.whl
```
yarikoptic, line 92: potentially auth could be replaced by using "trusted repositories" setup on pypi.

.github/workflows/pypi-whl.yml
```diff
@@ -0,0 +1,96 @@
...
+            ACTIVATE_PATH=.venv/bin/activate
+            BIDS_VALIDATOR_BINARY_NAME=bids-validator
+            VENV_BIN_DIR=.venv/bin
+          fi
+          source "$ACTIVATE_PATH"
+          pip install tools/pypi-whl/dist/*.whl
+          bids-validator --version
+          ls -lh "$VENV_BIN_DIR"
+          diff "$VENV_BIN_DIR"/"$BIDS_VALIDATOR_BINARY_NAME" tools/pypi-whl/dist/"$BIDS_VALIDATOR_BINARY_NAME"
+          deactivate
```
yarikoptic, line 77: move this into a helper script e.g. `tools/pypi-whl/test-installation` -- too much for the workflow - makes it hard to read.

.github/workflows/pypi-whl.yml
```diff
@@ -0,0 +1,96 @@
+name: PyPI Whl Publish
+
+on:
+  push:
+    branches:
+      - main
+      - deno-binary
```
yarikoptic, line 7: eventually remove 
```suggestion
```

yarikoptic, line 7: overall here aim for the same setup of `on` as in https://github.com/bids-standard/bids-validator/blob/main/.github/workflows/deno_tests.yml

```suggestion
on:
  push:
    branches: [main, dev]
    tags: ['*']
  pull_request:
    branches: [main, dev]
```

and then add corresponding `if` below for `twine` call:

.github/workflows/pypi-whl.yml
```diff
@@ -0,0 +1,96 @@
...
+          diff "$VENV_BIN_DIR"/"$BIDS_VALIDATOR_BINARY_NAME" tools/pypi-whl/dist/"$BIDS_VALIDATOR_BINARY_NAME"
+          deactivate
+      - name: Check if distribution's description will render correctly on PyPI
+        run: twine check tools/pypi-whl/dist/*.whl
+      - uses: actions/upload-artifact@v4
+        with:
+          name: wheel-${{ matrix.os }}
+          path: tools/pypi-whl/dist/*.whl
+      - name: Publish wheels if corresponds to a tag
+        # TODO: might be suboptimal since we might skip a tag if not the latest commit
```
yarikoptic, line 85:
```suggestion
        if: github.ref_type == 'tag'
```

.github/workflows/pypi-whl.yml
yarikoptic, line 92: with above `if` we likely do not need this 

```suggestion
            twine upload --verbose --repository testpypi dist/*.whl
```

effigies:
image

The `cp3*` is going to cause problems, as the package will only be installable with Python 3.9. Installing from test.pypi.org works because it finds `bids_validator_deno-2.0.7-py3-none-manylinux_2_17_x86_64.whl`.

You need to get back to the `py3-none-*`. I'm not sure how to do that in `pdm`. I could work it out in setuptools if I knew how to package a pre-built binary.

yarikoptic:
right, somehow that got lost I guess -- may be pdm is not executed within the folder where that pdb_build.py is?

Please add testing that whl files have `-py3-none-` in their names!

effigies:
Oh, I see. It reports using `cp39-cp39`, but then post-processing renames to `py3-none`.

candleindark:
I tried to use the pdm frontend to invoke the build locally. It didn't work for me. Just use `python -m build` if want to try it locally, but the current setup is really for CI to do the building.

.github/workflows/pypi-whl.yml
candleindark: done

.github/workflows/pypi-whl.yml
```diff
@@ -0,0 +1,80 @@
...
+        with:
+          name: wheel-${{ matrix.os }}
+          path: tools/pypi-whl/dist/*.whl
+      - name: Publish wheels if corresponds to a version tag
+        if: github.ref_type == 'tag'
+        working-directory: ../../
+        run: |
+          # Check if the tag is a version tag
+          if jq -e ".version[:1] != \"v\" and .version==\"$TAG\"" deno.json; then 
+              twine upload --verbose --repository testpypi dist/*.whl
```
yarikoptic, line 73: for the final version should remove 
```suggestion
              twine upload --verbose dist/*.whl
```

effigies:
Played with this a bit and wasn't entirely satisfied. Ended up spending way too much time on it. See #186.

PR #188, opened 2025-03-24
Add scrive for #180 and workflow to check for changes in changelog.d

rwblair:
Based on:
https://github.com/tarides/changelog-check-action/tree/main

### DIFF
--- a/changelog.d/20250324_121157_rosswilsonblair_scriv_for_180.md
+++ b/changelog.d/20250324_121157_rosswilsonblair_scriv_for_180.md
@@ -1,0 +1,46 @@
+<!--
+A new scriv changelog fragment.
+
+Uncomment the section that is right (remove the HTML comment wrapper).
+-->
+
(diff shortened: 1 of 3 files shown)


changelog.d/20250324_121157_rosswilsonblair_scriv_for_180.md
```diff
@@ -0,0 +1,46 @@
+<!--
+A new scriv changelog fragment.
+
+Uncomment the section that is right (remove the HTML comment wrapper).
+-->
+
+### Added
+
+- Create contexts for directories and validate their names according to schema.rules.directories entries.
```
effigies, line 9: IMO we should focus on user-facing features, while "contexts" are a little weedsy.

```suggestion
- Validate directory names names according to `schema.rules.directories` entries.
  Use rules to identify "opaque" directories whose contents are unspecified by BIDS.
```

PR #195, opened 2025-04-10
expand types for subject metadata in summary. Account for 89+ string also.

### LINKED ISSUE #194: `--json` option produces validation result inconsistent with type annotation
As of version 2.0.5, the output validation results produced with the `--json` option can have `SubjectMetadata` objects inconsistent with the type annotations in https://github.com/bids-standard/bids-validator/blob/39eff6f4ae8c5d9d0b1ec10a0d80fcd1b698f257/src/types/validation-result.ts#L3-L7 when validating some of the datasets in https://github.com/bids-standard/bids-examples and https://github.com/bids-standard/bids-error-examples. The inconsistency centers around the `age` and `sex` attributes of the objects. In the results, the `age` and `sex` attribute can sometimes be `null` or missing.

For example, when validating some of the datasets in the bids-examples repo,

```shell
bids-validator-deno --json --outfile out.json ieeg_epilepsyNWB
``` 
would produce a result containing a `subjectMetadata` object as
```json
        "subjectMetadata": [
            {
                "participantId": "01",
                "age": null,
                "sex": "m"
            }
        ]
```

and 

```shell
bids-validator-deno --json --outfile out.json micr_SEMzarr
```

would produce a result containing a `subjectMetadata` object as

```json
        "subjectMetadata": [
            {
                "participantId": "01",
                "sex": "M"
            }
        ]
```

when validating some of the datasets in bids-error-examples,

```shell
bids-validator-deno --json --outfile out.json invalid_pet001
```
would produce a result containing a `subjectMetadata` object as

```json
        "subjectMetadata": [
            {
                "participantId": "01"
            }
        ]
```

rwblair:
fixes #194

### DIFF
--- a/src/types/validation-result.ts
+++ b/src/types/validation-result.ts
@@ -1,7 +1,7 @@
 import type { DatasetIssues } from '../issues/datasetIssues.ts'
 
 export interface SubjectMetadata {
-  participantId?: string
+  participantId: string
   age?: number | null | "89+"
   sex?: string
 }
--- a/src/types/validation-result.ts
+++ b/src/types/validation-result.ts
@@ -1,9 +1,9 @@
 import type { DatasetIssues } from '../issues/datasetIssues.ts'
 
 export interface SubjectMetadata {
-  participantId: string
-  age: number
-  sex: string
+  participantId?: string
+  age?: number | null | "89+"
+  sex?: string
 }
 /*
     BodyPart: {},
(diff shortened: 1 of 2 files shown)


src/types/validation-result.ts
```diff
@@ -1,9 +1,9 @@
 import type { DatasetIssues } from '../issues/datasetIssues.ts'
 
 export interface SubjectMetadata {
-  participantId: string
-  age: number
-  sex: string
+  participantId?: string
```
effigies, line 4: Can this ever be empty?

```suggestion
  participantId: string
```

PR #197, opened 2025-04-14
Enh/read parquet files

rwblair:
Use hyparquet library to read read parquet files.Transforms them into a columns map, which may not be desirable. Parquet files have good typing metadata that is thrown away and this implementation, with everything becoming a string like when we read in TSV files. Type checking is then done via regex as per schema.objects.columns/schema.objects.formats entries.

Still need to implement maxRows.


rwblair:
After chat with @effigies, context.columns could become a union type with TSVs still using the old ColumnMap type, and parquets using whatever representation we choose for them. In applyRules validation we can rely on context extension to decide how to validate columns field. Parquet validation could then convert its known built in types to schema types to check against the schema.objects.columns types. This would get around type checking every value in a given column like we do for TSVs (and using regexes to do that type checking).

PR #201, opened 2025-05-02
Use schema.meta.associations during buildAssociations.

rwblair:
I have abused ts-expect-error, this should probably be remedied.

associations.ts doesn't have any direct tests, this should also probably be remedied.


changelog.d/20250502_103250_rosswilsonblair_use_association_selectors.md
```diff
@@ -0,0 +1,47 @@
...
+
+<!--
+### Added
+
+- A bullet item for the Added category.
+
+-->
+### Changed
+
+- Rely on schema.meta.associations to load context associations instead of relying on list maintained in validator.
```
effigies, line 16:
```suggestion
- Rely on `schema.meta.associations` to load context associations instead of relying on list maintained in validator.
```

effigies (approved):
LGTM. Small suggestion on the changelog.

PR #220, opened 2025-08-05
Bump hed-validator dependency to 4.1.0

happy5214:
This version includes multiple changes. While most relate to the new browser validation functionality and don't directly affect BIDS, some changes affecting BIDS include TypeScript type declarations and general bug fixes.


happy5214:
We'll have to have to fix the type errors (some are easier, some are harder) and issue a new release. Closing for now.

happy5214:
The type issues should be fixed now. I had two dataset issues locally, but they appeared to be unrelated to HED.

happy5214:
I give up. Unless you have any more ideas, I'm content to freeze the dependency at 4.0.1 until we port the validator to TypeScript.

effigies (approved):
Cool, I was able to undo the modifications to `build.ts`, so thanks for figuring this all out on your end. This was definitely beyond me.

PR #225, opened 2025-08-08
Raise error when json files are parsed and their root value is an array instead of object.

### LINKED ISSUE #178: Validate that a `.json` file contains an object
https://openneuro.org/datasets/ds004584/versions/1.0.0/file-display/task-Rest_events.json contains `[]`. The validator should check that JSON files actually are objects and not other JSON-parseable structures.

rwblair:
fixes #178

This is being thrown on sidecar loading so could generate a large number of errors for an inherited file.

### DIFF
--- a/changelog.d/20250808_135941_rosswilsonblair_raise_error_on_non_object_json.md
+++ b/changelog.d/20250808_135941_rosswilsonblair_raise_error_on_non_object_json.md
@@ -13,7 +13,7 @@ For top level release notes, leave all the headers commented out.
 -->
 ### Changed
 
-- Raise error when json files are parsed and their root value is an array instead of object
+- Raise error when JSON files are parsed and their root value is anything other than an object
 
 <!--
 ### Fixed
--- a/changelog.d/20250808_135941_rosswilsonblair_raise_error_on_non_object_json.md
+++ b/changelog.d/20250808_135941_rosswilsonblair_raise_error_on_non_object_json.md
@@ -1,0 +1,47 @@
+<!--
+A new scriv changelog fragment.
+
+Uncomment the section that is right (remove the HTML comment wrapper).
+For top level release notes, leave all the headers commented out.
+-->
+
+<!--
+### Added
+
+- A bullet item for the Added category.
+
+-->
+### Changed
+
+- Raise error when json files are parsed and their root value is an array instead of object
+
+<!--
+### Fixed
+
+- A bullet item for the Fixed category.
+
+-->
+<!--
+### Deprecated
+
+- A bullet item for the Deprecated category.
+
+-->
+<!--
+### Removed
+
+- A bullet item for the Removed category.
+
+-->
+<!--
+### Security
+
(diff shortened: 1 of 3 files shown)


src/files/json.ts
```diff
@@ -22,9 +22,14 @@ async function readJSONText(file: BIDSFile): Promise<string> {
...
+  let parsedText;
   try {
-    return JSON.parse(text)
+    parsedText = JSON.parse(text)
   } catch (error) {
     throw { key: 'JSON_INVALID' } // Raise syntax errors
   }
+  if (Array.isArray(parsedText)) {
+    throw { key: 'JSON_NOT_AN_OBJECT' }
+  }
```
effigies, line 33: It could also be a number or string. We want to check that it is actually an object, I think.

changelog.d/20250808_135941_rosswilsonblair_raise_error_on_non_object_json.md
```diff
@@ -0,0 +1,47 @@
...
+
+<!--
+### Added
+
+- A bullet item for the Added category.
+
+-->
+### Changed
+
+- Raise error when json files are parsed and their root value is an array instead of object
```
effigies, line 16:
```suggestion
- Raise error when JSON files are parsed and their root value is anything other than an object
```

src/issues/list.ts
```diff
@@ -9,6 +9,10 @@ export const bidsIssues: IssueDefinitionRecord = {
     severity: 'error',
     reason: 'Not a valid JSON file.',
   },
+  JSON_NOT_AN_OBJECT: {
+    severity: 'error',
+    reason: 'Parsed json file contains an array as its root element and not an object.',
```
effigies, line 14:
```suggestion
    reason: 'Parsed JSON file does not contain an object.',
```

src/files/json.ts
```diff
@@ -22,9 +22,14 @@ async function readJSONText(file: BIDSFile): Promise<string> {
...
   const text = await readJSONText(file) // Raise encoding errors
+  let parsedText;
   try {
-    return JSON.parse(text)
+    parsedText = JSON.parse(text)
   } catch (error) {
     throw { key: 'JSON_INVALID' } // Raise syntax errors
   }
+  if (Array.isArray(parsedText) || typeof parsedText !== "object") {
+    throw { key: 'JSON_NOT_AN_OBJECT' }
```
effigies, line 32: Would be good to show what we found. I think we use `evidence`?

```suggestion
    throw { key: 'JSON_NOT_AN_OBJECT', evidence: text.substring(0, 10) + (text.length > 10 ? '...' : '') }
```

effigies (approved):

src/files/json.ts
rwblair: Personally I'd still end up cat'ing the file if I received the error. But no harm in returning evidence.  `decoder.decode` used by readJSONText should always return a string so should be safe there.

PR #234, opened 2025-08-25
feat: Unify schema and sidecar-defined column validation

effigies:
This is a companion to https://github.com/bids-standard/bids-specification/pull/2162.

The strategy here is to map both the JSON-schema style definition and the sidecar definitions onto a common validation specification:

```ts
interface ValueSpec {
  pattern: RegExp
  levels?: string[]
  maximum?: number
  minimum?: number
}
```

This is sufficient to validate any values. For the purposes of comparing two definitions, we need a slightly larger signature:

```ts
interface ValueSignature {
  formats: string[]  // Array because anyOf
  pattern?: string
  units?: string
  levels?: string[]
  maximum?: number
  minimum?: number
}
```

The rest flows more-or-less naturally from there. Note that with `ValueSignature`, we can identify trivial specs that will always return `True` and just skip validating those columns.

### DIFF
--- a/changelog.d/20250824_210257_markiewicz_sidecar_columns.md
+++ b/changelog.d/20250824_210257_markiewicz_sidecar_columns.md
@@ -1,0 +1,48 @@
+<!--
+A new scriv changelog fragment.
+
+Uncomment the section that is right (remove the HTML comment wrapper).
+For top level release notes, leave all the headers commented out.
+-->
+
+<!--
+### Added
+
+- A bullet item for the Added category.
+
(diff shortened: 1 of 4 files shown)


effigies:
@nellh This PR is important to have before the spec release. Happy to talk it over at the schema sprint if it's not easily understandable.

nellh (approved):
This looks reasonable to me. Tested with some example datasets and matches the expected behavior in bids-standard/bids-specification#2162.

PR #242, opened 2025-08-28
feat: Split out deprecation warning for age=="89+"

effigies:
This allows 89+ to be used in age columns, regardless of whether we start raising errors for optional columns.


effigies:
@nellh LMK if you think this should go in `main` or `dev`.

effigies:
Added a test. Will merge on pass.

Post-merge reviews welcome.

PR #247, opened 2025-08-29
feat: Always error on incorrect TSV types

effigies:
Following #234, it is always possible to define a type precisely in a sidecar definition. It is therefore safe to error when types don't match, because the solution is "Improve the type of your column."

Builds on #240


effigies:
As discussed with @nellh. The change is simple enough that I'm charging ahead.

It can be backed out in a bugfix release if others disagree strongly.

PR #243, opened 2025-08-29
feat: Allow initial and index columns to be non-required

effigies:
Optional initial columns are necessary for BEPs 20 and 42. Optional index columns are necessary for BEP 42.

PR #254, opened 2025-09-09
rf: Validate additional columns when defined

### LINKED ISSUE #250: Validate dataset-defined columns
Currently, columns that are defined in JSON sidecars are not validated at all, because we are only checking datasets the the columns of `schema.rules.tabular_data` rules.

effigies:
This refactor made it easier enough to make `evalAdditionalColumns` one `if` branch in `evalColumns`.

This will trigger some examples to fail.

Closes #250.


src/schema/tables.ts
```diff
@@ -409,7 +380,7 @@ export function evalIndexColumns(
   const uniqueIndexValues = new Set()
   const index_columns = rule.index_columns.map((col: string) => {
     return schema.objects.columns[col].name
-  }).filter((col: string) => col in context.columns)
+  }).filter((col: string) => context.columns[col])
```
effigies, line 383: This was a bug introduced in #243. Sorry.

PR #260, opened 2025-09-12
BEP020 patches

effigies:
TODO:

* ~~Message interpolation~~ See #267 

xref

* https://github.com/bids-standard/bids-specification/pull/1128
* https://github.com/bids-standard/bids-examples/pull/514

PR #262, opened 2025-09-15
feat: Cache JSON file loads

effigies:
Because loading associated file sidecars is going to be necessary for #260, it makes sense to start caching JSON loads.


src/files/json.test.ts
```diff
@@ -18,9 +19,12 @@ function encodeUTF16(text: string) {
   return buffer
 }
 
-function makeFile(text: string, encoding: string): BIDSFile {
+function makeFile(path: string, text: string, encoding: string): BIDSFile {
```
rwblair, line 22: Surprised we got away with not having path here for so long.

PR #265, opened 2025-09-16
Add per context check to see if any files in its directory have the same name with different case

### LINKED ISSUE #49: add check for case collision intollerance
original proposal as a PR in BIDS: https://github.com/bids-standard/bids-specification/pull/858 seems to be getting approved and if merged soon will enter a release after v1.6.0-203-g811e17e2 ,

rwblair:
This implementation uses toLowerCase, the other option would be localeCompare. Folk wisdom online being localeCompare is slower, but works better for more languages.

fixes #49


effigies (approved):
LGTM. Left a small suggestion, which you should feel free to take or leave.

PR #268, opened 2025-09-17
BEP042 patches

drammock:
This is needed for BEP042, specifically for [bids-standard/bids-specification@`5ea4b08` (#1998)](https://github.com/bids-standard/bids-specification/pull/1998/commits/5ea4b086534b5e17f55c8d0d82a017f40af7ab88)

~~Code is copied (with minor typo fixes) from [this forum answer](https://neurostars.org/t/how-to-write-complex-validator-rules-checks/34125/5) so I've listed @rwblair in co-authored-by~~

~~marking as draft because I'm not 100% sure that just `keys` will allow us to check everything we want to check.~~

*EDIT: approach changed after real-time convo with @effigies; now this collates multiple coordsystem files*

xref

* https://github.com/bids-standard/bids-specification/pull/1998
* https://github.com/bids-standard/bids-examples/pull/480

### DIFF
--- a/changelog.d/20250919_172910_dan_add_keys_to_assoc_coordsys.md
+++ b/changelog.d/20250919_172910_dan_add_keys_to_assoc_coordsys.md
@@ -1,0 +1,48 @@
+<!--
+A new scriv changelog fragment.
+
+Uncomment the section that is right (remove the HTML comment wrapper).
+For top level release notes, leave all the headers commented out.
+-->
+
+
+### Added
+
+- New `associations.coordsystems` to collate coordsystem files.
+
+<!--
+### Changed
+
+- A bullet item for the Changed category.
+
+-->
+<!--
+### Fixed
+
+- A bullet item for the Fixed category.
+
+-->
+<!--
+### Deprecated
+
+- A bullet item for the Deprecated category.
+
+-->
+<!--
+### Removed
+
+- A bullet item for the Removed category.
+
+-->
+<!--
+### Security
+
+- A bullet item for the Security category.
+
+-->
+<!--
+### Infrastructure
+
+- A bullet item for the Infrastructure category.
+
+-->
--- a/changelog.d/20250919_172910_dan_add_keys_to_assoc_coordsys.md
+++ b/changelog.d/20250919_172910_dan_add_keys_to_assoc_coordsys.md
@@ -8,7 +8,8 @@ For top level release notes, leave all the headers commented out.
 
 ### Added
 
-- New `associations.coordsystems` to collate coordsystem files.
+- Implement `associations.coordsystems` to collate multiple `coordsystem.json` files,
+  as required by BEP 042 (EMG).
 
 <!--
 ### Changed
--- a/src/schema/associations.ts
+++ b/src/schema/associations.ts
@@ -111,7 +111,7 @@ const multiAssociationLookup: Record<string, MultiLoadFunction> = {
   coordsystems: async (
     files: BIDSFile[],
     options: any,
-  ): Promise<{ paths: string[]; spaces: string[]; parents: string[] }> => {
+  ): Promise<{ paths: string[]; spaces: string[]; ParentCoordinateSystems: string[] }> => {
     const jsons = await Promise.allSettled(
       files.map((f) => loadJSON(f).catch(() => ({} as Record<string, unknown>))),
     )
--- a/src/schema/associations.ts
+++ b/src/schema/associations.ts
@@ -9,12 +9,15 @@ import { parseBvalBvec } from '../files/dwi.ts'
 import { readSidecars, walkBack } from '../files/inheritance.ts'
 import { evalCheck } from './applyRules.ts'
 import { expressionFunctions } from './expressionLanguage.ts'
+import { readEntities } from './entities.ts'
 
 import { readText } from '../files/access.ts'
 
 interface WithSidecar {
   sidecar: Record<string, unknown>
 }
+type LoadFunction = (file: BIDSFile, options: any) => Promise<any>
+type MultiLoadFunction = (files: BIDSFile[], options: any) => Promise<any>
 
 function defaultAssociation(file: BIDSFile, _options: any): Promise<{ path: string }> {
   return Promise.resolve({ path: file.path })
@@ -34,7 +37,7 @@ async function constructSidecar(file: BIDSFile): Promise<Record<string, unknown>
  *
  * Many associations only consist of a path; this object is for more complex associations.
  */
-const associationLookup = {
+const associationLookup: Record<string, LoadFunction> = {
   events: async (file: BIDSFile, options: { maxRows: number }): Promise<Events & WithSidecar> => {
     const columns = await loadTSV(file, options.maxRows)
       .catch((e) => {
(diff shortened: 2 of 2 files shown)


src/schema/associations.ts
```diff
@@ -85,6 +86,16 @@ const associationLookup = {
       sampling_frequency: columns.get('sampling_frequency'),
     }
   },
+  coordsystem: async (file: BIDSFile, options: { maxRows: number }): Promise<Channels> => {
+    const keys = Object.keys(await loadJSON(file, options.maxRows)
+      .catch((e) => {
+        return new Map()
```
effigies, line 92:
```suggestion
        return []
```

src/schema/associations.ts
```diff
@@ -85,6 +86,16 @@ const associationLookup = {
       sampling_frequency: columns.get('sampling_frequency'),
     }
   },
+  coordsystem: async (file: BIDSFile, options: { maxRows: number }): Promise<Channels> => {
```
effigies, line 89: This return type needs to match.


```suggestion
  coordsystem: async (file: BIDSFile, options: { maxRows: number }): Promise<{path: string, keys: string[]}> => {
```

drammock: heh. that's embarassing. can you tell I don't usually write TypeScript?

src/schema/associations.ts
```diff
@@ -85,6 +86,16 @@ const associationLookup = {
       sampling_frequency: columns.get('sampling_frequency'),
     }
   },
+  coordsystem: async (file: BIDSFile, options: { maxRows: number }): Promise<{path: string, keys: string[]}> => {
+    const keys = Object.keys(await loadJSON(file, options.maxRows)
+      .catch((e) => {
+        return []
+      }))
```
effigies, line 93: Realized that the await was inside Object.keys. Also, there's no maxrows. We should add a test to make sure the file gets loaded. I think it wouldn't with the extra argument, and nothing to notice the bug.

```suggestion
    const keys = Object.keys(await loadJSON(file).catch((e) => {}))
```

drammock: FWIW the file *was* getting loaded in my local tests (based on a console.log of the context).

effigies: Huh. Javascript.

drammock:
> Notice that `associations` is empty.

looks like you resolved it on the spec end. I guess this is ready for changelog and review/merge then?

changelog.d/20250919_172910_dan_add_keys_to_assoc_coordsys.md
```diff
@@ -0,0 +1,48 @@
...
+A new scriv changelog fragment.
+
+Uncomment the section that is right (remove the HTML comment wrapper).
+For top level release notes, leave all the headers commented out.
+-->
+
+
+### Added
+
+- New `associations.coordsystems` to collate coordsystem files.
```
effigies, line 11:
```suggestion
- Implement `associations.coordsystems` to collate multiple `coordsystem.json` files,
  as required by BEP 042 (EMG).
```

src/schema/context.ts
```diff
@@ -162,6 +163,9 @@ export class BIDSContext implements Context {
     this.columns = new ColumnsMap() as Record<string, string[]>
     this.json = {}
     this.associations = {} as Associations
+    if (this.dataset.schema.objects) {  // schema may be empty for some tests
+      datatypeFromDirectory(this.dataset.schema, this)
+    }
```
effigies, line 168: @rwblair This is necessary to allow associations to use `datatype` or `modality` in their selectors. I would like to refactor datatype/modality handling, as that really doesn't need to be part of `filenameIdentify` stages. That will be a separate PR, when I have the time.

I'm realizing now that our existing policy of clearing out some path metadata if file rules fail to match will mean that association selectors get different results from, e.g., sidecar or check selectors. This may need further thinking through.

src/schema/associations.ts
```diff
@@ -86,6 +91,24 @@ const associationLookup = {
     }
   },
 }
+const multiAssociationLookup: Record<string, MultiLoadFunction> = {
```
effigies, line 94: It was easier for me to wrangle the type system by adding a second lookup table than by trying to make `associationLookup` hold two type signatures.

src/files/inheritance.ts
```diff
@@ -50,7 +50,7 @@ export function* walkBack<T extends string[]>(
     if (candidates.length > 1) {
       const exactMatch = candidates.find((file) => {
         const { entities } = readEntities(file.name)
-        return Object.keys(sourceParts.entities).every((entity) =>
+        return [...Object.keys(sourceParts.entities), ...(targetEntities ?? [])].every((entity) =>
```
effigies, line 53: This is a logic bug that doesn't matter in the current operations because we've only cared to get the first value, but what was happening was that an `ieeg.*` file with multiple `space-*_electrodes.tsv` files would just match the first one because every entity in `ieeg` was also in `electrodes`, and checking that `entities['space'] === sourceEntities['space']` wasn't happening.

It should be fixed and tested in a separate PR, but the bug needed fixing to make this work.

src/schema/associations.ts
```diff
@@ -123,33 +146,37 @@ export async function buildAssociations(
...
-      if (
-        error && typeof error === 'object' && 'code' in error &&
-        error.code === 'MULTIPLE_INHERITABLE_FILES'
-      ) {
-        // @ts-expect-error
+    } catch (error: any) {
+      if (error?.code === 'MULTIPLE_INHERITABLE_FILES') {
         context.dataset.issues.add(error)
-        break
+        continue
```
effigies, line 152: I *think* we didn't really want to stop building associations on the first one that found `MULTIPLE_INHERITABLE_FILES`. With the `break`, the order of loading associations changes the behavior of the validator; with `continue`, it should be consistent, regardless.

This can also be a separate PR.

src/schema/associations.ts
```diff
@@ -123,33 +146,37 @@ export async function buildAssociations(
         rule.target.suffix,
         rule.target?.entities ?? [],
       ).next().value
```
effigies, line 148: The EMG examples do not have this case, but with multiple `coordsystem.json`s, you could imagine having some at the root and some at the leaf.

We currently strongly assume (and implement with `.next().value`) that the first level found is the level of the association, which has worked for single-file associations. This probably needs to be relaxed, but the logic will be more intricate. I think you would need to take the first `coordsystem.json` with a given `space` entity, not the full collection.

For now this could be noted as a limitation in the validator implementation.

src/schema/associations.ts
```diff
@@ -123,33 +146,37 @@ export async function buildAssociations(
...
-        },
-      )
+    if (file && !(Array.isArray(file) && file.length === 0)) {
+      const options = { maxRows: context.dataset.options?.maxRows }
+      if (key in multiAssociationLookup) {
+        const load = multiAssociationLookup[key]
+        if (!Array.isArray(file)) {
+          file = [file]
+        }
+        associations[key as keyof Associations] = await load(file, options)
```
effigies, line 165: Didn't include a `catch` since `location` is not obvious. Might just need to silently swallow and hope the error shows up when loading the `coordsystem.json` file on its own?, e.g.,

```suggestion
        associations[key as keyof Associations] = await load(file, options).catch((e: any) => {})
```

src/schema/associations.ts
```diff
@@ -123,33 +146,37 @@ export async function buildAssociations(
...
+      } else {
+        const load = associationLookup[key] ?? defaultAssociation
+        if (Array.isArray(file)) {
+          file = file[0]
+        }
+        const location = file.path
+        associations[key as keyof Associations] = await load(file, { maxRows: context.dataset.options?.maxRows }).catch(
```
effigies, line 172: Meant to do this:

```suggestion
        associations[key as keyof Associations] = await load(file, options).catch(
```

effigies (commented):
Leaving some notes. This PR can be merged once @rwblair is happy with it, or remain open during the BEP42 review period. It would need to be merged (and the `dev` branch updated) before the examples PR can be merged.

effigies (approved):

PR #270, opened 2025-09-18
Throw an error, do not just log an error, if provided schema is not legit

yarikoptic:
I think current behavior violates the principle of the least surprise: I ran validation while providing custom schema source, it executed with many warnings and errors reported to the screen. It is only after scrolling far up, I saw that actually my specification was wrong and operation resorted to built-in schema.

Moreover, having varying results depending on network flukes is also not a good behavior.

TODOs (if agreed in principle)
- [x] add changelog
- [ ] fix or add a test (there should be a test for such behavior)

### DIFF
--- a/changelog.d/20250918_173356_yarikoptic_enh_crash.md
+++ b/changelog.d/20250918_173356_yarikoptic_enh_crash.md
@@ -1,0 +1,4 @@
+### Changed
+
+- Throw an error if specified schema (e.g. via `--schema` or `BIDS_SCHEMA` env
+  var) could not be loaded.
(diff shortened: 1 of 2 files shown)


effigies (approved):
I've been bitten by this. I can't remember why we decided this was a good idea, but I disagree with past us.

effigies:
What do you think about the case where someone has BIDS_SCHEMA in their environment but doesn't know it. Should we indicate where the URL came from?

yarikoptic:
that might come useful although I do not think any other software I know bothers to tell apart where config value came from.

If to add -- do you feel it is better as error post-processing (check again if env var available) or pair with

- #269 

and add some `schemaSourceMethod = CLIOption|envvar`  (feels overweight FWIW)?

PR #271, opened 2025-09-19
feat: add --format option for output formatting

### LINKED ISSUE #251: add --pp-json - --json with pretty print
it is frequent desire to use/see produced .json by a human. ATM it requires additional piping through `jq` or alike.

I think it might be beneficial, especially for use with `--outfile` to have `--pp-json` mode.  

But I am about to suggest one more mode ( ;) ) so I wonder if it should altogether RFed into `-f|--output-format text| json| json_pp` to

- facilitate possible expansion if needed
- might be cleaner in the code internally ;-)

djken2009:
- Add --format option with support for text, json, and json_pp formats
- json_pp provides pretty-printed JSON output for human readability
- Maintain backward compatibility with existing --json option
- Resolves #251

### DIFF
--- a/src/setup/options.test.ts
+++ b/src/setup/options.test.ts
@@ -12,6 +12,7 @@ Deno.test('options parsing', async (t) => {
       blacklistModalities: [],
       datasetTypes: [],
       maxRows: 1000,
+      format: 'text',
     })
   })
 })


effigies (approved):
I broadly support the proposal, will give people time to weigh in on the color of this bikeshed.

PR #279, opened 2025-10-03
fix: Weakly reference parent FileTrees, upgrade hed-validator

### LINKED ISSUE #278: FileTree and BIDSFile contain circular references
The `FileTree` and `BIDSFile` classes refer to each other circularly. I believe this may be the real root cause of the issues that precipitated #232 and https://github.com/hed-standard/hed-javascript/issues/586, as the current code also crashes on certain datasets, e.g. https://github.com/hed-standard/hed-examples/tree/39cfc99aeaca3afb2c43bef9b1b6d52b3321ff08/datasets/fmri_soccer21s_hed. The circular references need to be augmented with `WeakRef`, which requires additional code changes.

effigies:
I've confirmed that this no longer hangs:

```console
$ deno run -A src/bids-validator.ts --ignoreWarnings tests/data/bids-examples/synthetic
```

Closes #278.

### DIFF
--- a/changelog.d/20251003_131902_markiewicz_weakref_parents.md
+++ b/changelog.d/20251003_131902_markiewicz_weakref_parents.md
@@ -1,0 +1,5 @@
+### Fixed
+
+- Circular references that could lead to a hanging process were mitigated. [#278]
+
+[#278]: https://github.com/bids-standard/bids-validator/issues/278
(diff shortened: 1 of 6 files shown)


happy5214:
Can you not just keep the constructors the same and co-opt the setters to construct the initial `WeakRef`s?

rwblair (approved):

PR #282, opened 2025-10-14
feat: Start loading contexts earlier, buffer contexts for better concurrency

effigies:
This PR has two pieces, neither of which produce any noticeable speedup on their own. The first creates a `context.loaded` Promise that indicates whether the context is fully loaded. The second creates a queue that can hold onto initializing contexts until we're ready for them.

For ds000001, I was getting about 9s per run. With a queue of 5, down to 4s, and a queue of 20 got us down to 2s.


effigies:
Okay, reworked this to perform `new BIDSContext()` and `await context.loaded` in `walkFileTree` or in a helper `async function makeBIDSContext()`. As long as nobody calls `new BIDSContext()` directly instead of through these, the resources should get released.

PR #297, opened 2025-11-24
fix: Set nifti_header.axis_codes to null for bad qforms, rather than error

effigies:
In #112 we added `axis_codes` to the `nifti_header` context, which unwittingly made the whole header appear unreadable for improper quaternions (the norm of the quaternion should be 1; if the norm of the explicit components exceeds 1, the implicit component can't be calculated). Raising an error in this case is not worth it, since it only enables a warning check.

I've seen two reports in the last week about surprising `NIFTI_HEADER_UNREADABLE` errors (OpenNeuro support ticket and https://neurostars.org/t/bids-validator-error-nifti-header-unreadable/34423), which result from this.

### DIFF
--- a/changelog.d/20251124_080614_markiewicz_bad_qform.md
+++ b/changelog.d/20251124_080614_markiewicz_bad_qform.md
@@ -1,0 +1,5 @@
+### Fixed
+
+- NIfTI files with bad qform matrices, resulting from non-normalized quaternions,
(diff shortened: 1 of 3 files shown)


rwblair (approved):

effigies:
Just a note that there's a related upstream fix: https://github.com/rii-mango/NIFTI-Reader-JS/pull/56

PR #329, opened 2025-12-18
fix: use .bids-validator-config.json by default if available.

### LINKED ISSUE #293: Return default config file to be .bids-validator-config.json ?
Used to be the case in legacy

Image

ATM current one does have `--config` but no default:

```
❯ uvx bids-validator-deno --help | grep config

  -c, --config           <file>          - Path to a JSON configuration file                               
❯ uvx bids-validator-deno --version

bids-validator 2.1.0
```

rwblair:
fixes #293 

Sins committed in this PR:

- Web version iterates through all files from directory upload to find .bids-validator-config.json
- Web versions uses `alert()`


src/main.ts
```diff
@@ -28,7 +29,21 @@ export async function main(): Promise<ValidationResult> {
...
+       const defaultConfig = join(absolutePath, '.bids-validator-config.json') 
+    try {
+      await Deno.lstat(defaultConfig)
+       config = JSON.parse(Deno.readTextFileSync(defaultConfig))
+       options.config = defaultConfig
+    } catch {
+      if (!(err instanceof Deno.errors.NotFound)) {
+        throw err;
+      }
+    }
```
effigies: What if we did what we do for `.bidsignore` and pull it out of the `tree`? We don't need a filesystem hit to check if it's there.

```suggestion
    const configFile = tree.get('.bids-validator-config.json') as BIDSFile
    if (configFile) {
      config = await configFile.text().then((text) => JSON.parse(text))
    }
```

We could also use our JSON loader and create an issue. Probably better not to crash...

web/src/App.tsx
```diff
@@ -50,11 +50,19 @@ function App() {
...
       recursive: true,
     })
+    let config = {}
+    const configFile = dirHandle.find(file => file.name ==='.bids-validator-config.json') 
+    if (configFile) {
+      config = configFile.text().then(text => JSON.parse(text)).catch((err) => {
+        alert(`Failed to load ".bids-validator-config.json". \n\nUsing empty configuration object:\n\n${err}`)
+      })
+    }
     const fileTree = await fileListToTree(dirHandle)
```
effigies: Similarly:

```suggestion
    const fileTree = await fileListToTree(dirHandle)
    let config = {}
    const configFile = fileTree.get('.bids-validator-config.json') as BIDSFile
    if (configFile) {
      config = await configFile.text().then(text => JSON.parse(text)).catch((err) => {
        alert(`Failed to load ".bids-validator-config.json". \n\nUsing empty configuration object:\n\n${err}`)
      })
    }
```

effigies:
Oh, and we probably need to add `.bids-validator-config.json` to our default ignores.

src/main.ts
rwblair: Using `loadJSON` for the default config and std functions for the cli specified config cause the error objects produced to be different.

The crash from either error is uninformative as currently implemented. Injecting these caught errors as issues into the dataset issues feels wrong for two reasons. It isn't an issue with validity with respect to the specification and its likely to get lost the sea of errors/issues. Treating them as issues instead of console output is good for people who consume the json validation results. Should we be outputting logger messages in the json output object? 

We could treat them as a separate class of issue that gets special treatment with their own section in validation results. For standard output these could be printed immediately before or after the summary output to highlight them since we are changing the behavior of the validation from what was likely intended by the user as opposed to reporting a normal validation issue. The failure to load a user specified schema would fall under this new class of issue. Currently we log some output saying that we are falling back to the default schema but this output is at the beginning of validation and is hard to notice, its something that I've been bit by before.

rwblair:
> Oh, and we probably need to add `.bids-validator-config.json` to our default ignores.

All dot files are currently included by default.

src/main.ts
rwblair: [cdc611c](https://github.com/bids-standard/bids-validator/pull/329/commits/cdc611ca1395d09887b9c3f85d943cc557fad767) Wraps the both json loads in a try catch that calls logger.error, this is at least closer to the behavior of the web validator as is. I'm still open to making them issues.

Example output with config file specified:
```
ERROR 
      Failed to load /home/rwblair/projects/bids/bids-examples/ds001/.bids-validator-config.json using empty configuration object.
      SyntaxError: Unexpected end of JSON input
    at JSON.parse (<anonymous>)
    at main (file:///home/rwblair/projects/bids/bids-validator/src/main.ts:37:21)
    at eventLoopTick (ext:core/01_core.js:179:7)
    at async file:///home/rwblair/projects/bids/bids-validator/src/bids-validator.ts:4:16
```
    
For an empty implicit config file:
```
      Failed to load .bids-validator-config.json using empty configuration object.
      {"code":"JSON_INVALID"}
```

rwblair: Here's an implementation that doesn't use stat, has both types of config files produce errors of the same type, and catches the error:
```
  let config = {}
  const configFile = options.config ?? join(absolutePath, '.bids-validator-config.json')
  config = await Deno.readTextFile(configFile).then((configText) => JSON.parse(configText)).catch(
    (error) => {
      if (options.config || !(error instanceof Deno.errors.NotFound)) {
        logger.error(`
          Failed to load ${configFile son'}. Using empty configuration object.
          ${error.stack || JSON.stringify(error)}
        `)
      }
    },
  )
```

PR #333, opened 2026-01-12
feat: Presign S3 URLs if AWS credentials are detected

effigies:
Addresses https://github.com/OpenNeuroOrg/openneuro/issues/3694.


nellh (approved):
This works well, it does run into an issue with memoize on readRemotes causing a memory leak. I'll submit a fix for that separately.

PR #335, opened 2026-01-13
fix: Add memoize resolver option and default to serialization of all arguments

nellh:
While reviewing #333 I ran into a memory leak where dangling references to the packfile object references were held by memoize cache keys. Avoiding using these objects as keys prevents the memory leak but it's probably safer to serialize by default and except anything where this is an issue. The resolver function can be used in any case where memoization should be a partial key of the arguments.

Limits the memoize key to the repo path for readRemotes to avoid serializing the full options object. This is also done for modalityTable to use the object as the key to skip serialization in that case as it is stable.

### DIFF
--- a/changelog.d/20260113_101504_nell_memoize_fixes.md
+++ b/changelog.d/20260113_101504_nell_memoize_fixes.md
@@ -1,0 +1,47 @@
+<!--
+A new scriv changelog fragment.
+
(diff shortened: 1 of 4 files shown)


effigies (approved):
LGTM. Thanks for catching this.

PR #338, opened 2026-01-15
feat: Extract AWS region from special remote, default to us-east-1

effigies:


src/files/repo.ts
```diff
@@ -205,14 +205,16 @@ export async function resolveAnnexedFile(
     // Take the newest remote (reverse sort timestamps, take first)
     uuid = Object.entries(rmet).toSorted((a, b) => +b[1].timestamp - +a[1].timestamp)[0][0]
   }
-  const { host, bucket } = remotes[uuid]
+  const { host, bucket, region } = remotes[uuid]
```
effigies:
```suggestion
  let { host, bucket, region } = remotes[uuid]
```

PR #351, opened 2026-02-02
fix: Add timeout to avoid resource leak, retry failed fetches

effigies:
Builds on #349.

### DIFF
--- a/src/files/access.ts
+++ b/src/files/access.ts
@@ -1,6 +1,7 @@
 import { type BIDSFile } from '../types/filetree.ts'
 import { type Issue } from '../types/issues.ts'
 import { filememoize } from '../utils/memoize.ts'
+import { logger } from '../utils/logger.ts'
 
 function IOErrorToIssue(err: { code: string; name: string }): Issue {
   let issueMessage: string | undefined = undefined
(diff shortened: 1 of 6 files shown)


src/files/openers.ts
```diff
@@ -107,12 +118,25 @@ export class HTTPOpener implements FileOpener {
...
+        logger.error(`Error during fetch retry check for ${this.url}: ${error}`)
+        return (
+          error instanceof TypeError ||
+          error instanceof HttpError && (error.status == 429 || error.status >= 500) ||
+          error instanceof DOMException && error.name === 'TimeoutError'
+        )
+      },
     })
+    return result
   }
```
effigies: AFAICT this `isRetriable` is never getting called. I don't really understand why.

src/files/access.ts
```diff
@@ -14,6 +15,7 @@ export async function openStream(
   file: BIDSFile,
 ): Promise<ReadableStream<Uint8Array<ArrayBuffer>>> {
   return file.stream().catch((err: any) => {
+    logger.critical(`Error opening stream from file ${file.path}: ${err}`)
```
effigies: For some reason, this `catch` is getting the `TimeoutError`s, but the one in retry isn't.

PR #366, opened 2026-03-16
Enh/bep036 extensions

rwblair:
Add zip and glob to expression language. 
Add dirEntities to file context.
Add root level session.tsv to dataset context.

Not all these may end up being needed depending on how the final rules shake out.


rwblair:
_walk call added to expressionlanguage.ts should be memoized separately.

Original glob implementation had split the pattern to match by `/` and would abandon testing branches of the file tree that failed as it walked. I couldn't figure out a clean way to make this work with `**` so I went for the knuckle dragging but compact seen here.

PR #370, opened 2026-03-26
feat: Add --git-ref option for git tree validation (bare repository support)

effigies:
This PR aims to add bare-repository support with both local and remote annexed files. This will also allow for passing a specific git reference to the validator so that prior versions can be re-validated without checking out a worktree.


src/files/git.ts
```diff
@@ -0,0 +1,351 @@
...
+      resolvedOid,
+      gitdir,
+      gitOptions,
+      symlinkMap,
+      preferredRemote,
+    )
+    if (opener) {
+      const file = new BIDSFile('/' + filepath, opener, ignore)
+      files.push(file)
+    }
```
effigies, line 347: We may want to create an issue for unresolveable symlinks. Given that we don't have an issues list yet, one way to do this would be an Opener that raises when accessed.

effigies (commented):
I've gone over this again and adjusted it. One final question I have is what to do with unresolveable symlinks. Right now we just exclude them from the file tree, but I think they should probably be an error. The only thing is that we don't have an equivalent consideration in the filesystem-based FileTree builder, so I think we want to be consistent between the two cases.

src/files/git.ts
rwblair: My first inclination is to add an array of issues to the core `FileTree` class? Proper Issue objects could be generated by different read tree functions and then the issues could be merged into the dataset context issues at a later time.

Right now if permissions of a subdirectory are messed up the entire readFileTree for deno will fail and stop execution. Catching these errors, making issues for them and continuing validation would make the validator a little bit more flexible.

effigies: After hemming and hawing back and forth with Claude, I think this needs discussion. I put together a summary document and we can talk this over: [Symlink Handling in bids-validator](https://bids-validator--370.org.readthedocs.build/en/370/dev/discussion/2026-04-symlink_handling.html)

effigies:
@rwblair Polished up the doc in 840b2b423babb4f5dcadbb98154ef1f274fde9b0. Want to have a look? Let's defer implementation of those decisions to another PR, but if you're good with this PR and the state of that doc, please merge.

Issue #373, opened 2026-04-07
singularity/apptainer can't run OCI image from docker


yarikoptic:
<details>
<summary>fwiw -- double checked that podman is ok</summary> 

```shell
$> podman run -it --rm bids/validator:2.4.1 --help
Usage:   bids-validator <dataset_directory>
Version: 2.4.1                             

Description:

  This tool checks if a dataset in a given directory is compatible with the Brain Imaging Data Structure specification. To learn more about Brain
  Imaging Data Structure visit http://bids.neuroimaging.io                                                                                       

Options:

  -h, --help                                - Show this help.                                                                                        
  -V, --version                             - Show the version number for this program.                                
```
</details>

but if trying via singularity, since image is read-only

```shell
$> singularity run docker://bids/validator:2.4.1
INFO:    Converting OCI blobs to SIF format
...
INFO:    Creating SIF file...
Could not initialize cache database '/deno-dir/dep_analysis_cache_v2', deleting and retrying... (SqliteFailure(Error { code: CannotOpen, extended_code: 14 }, Some("unable to open database file: /deno-dir/dep_analysis_cache_v2"
)))
Could not initialize cache database '/deno-dir/v8_code_cache_v2', deleting and retrying... (SqliteFailure(Error { code: CannotOpen, extended_code: 14 }, Some("unable to open database file: /deno-dir/v8_code_cache_v2")))
Failed to open cache file '/deno-dir/dep_analysis_cache_v2', opening in-memory cache.
Failed to open cache file '/deno-dir/v8_code_cache_v2', performance may be degraded.
Could not initialize cache database '/deno-dir/node_analysis_cache_v2', deleting and retrying... (SqliteFailure(Error { code: CannotOpen, extended_code: 14 }, Some("unable to open database file: /deno-dir/node_analysis_cache_v
2")))
Failed to open cache file '/deno-dir/node_analysis_cache_v2', opening in-memory cache.
error: Module not found "file:///tmp/bids-validator.js".
```

I would have hoped that the whole app would be "fully compiled" by then and not require any temp folders etc.  I have tried a workaround which avoids problem with deno-dir but

```
$> SINGULARITYENV_DENO_DIR=/tmp/.deno singularity run docker://bids/validator:2.4.1
INFO:    Using cached SIF image
error: Module not found "file:///tmp/bids-validator.js".
```

so there is more to it  :-/ 

originally reported by @FeHoff in

[...]

PR #375, opened 2026-04-07
Hardcode abs path to bids-validator.js for ENTRYPOINT in Dockerfile

yarikoptic:
that should allow for singularity to run such a container where its workdir differs from hardcoded in Dockerfile for docker `/src`

- addresses part of #373 , the other part might be worked around by pointing ` DENO_DIR=/tmp/deno` or alike (better uses mktemp!), but it is somewhat orthogonal here


effigies (approved):

effigies:
Can you add a quick changelog entry? `uvx scriv create --edit`

PR #376, opened 2026-04-09
rf/sty: Address linting issues

effigies:
`deno lint` output is too voluminous to address manually, so we haven't. I let Claude walk through the errors, have reviewed all of them, and it lints/tests clean. This should make `deno lint` an actually useful tool moving forward.

The main change of consequence here is that our check functions now can return `void | Promise<void>`, allowing synchronous functions to be synchronous. It's safe to `await` a synchronous function.

Some future work could be going through all of the `// @ts-*` comments and seeing what can actually be made type-safe.


src/files/repo.ts
```diff
@@ -66,10 +68,13 @@ const createS3Client = memoize(_createS3Client)
...
 export async function readAnnexPath(
   filepath: string,
-  options: any,
+  // deno-lint-ignore no-explicit-any
+  options: Record<string, any>,
 ): Promise<string> {
-  const oid = await git.resolveRef({ ref: 'git-annex', ...options })
-  const { blob } = await git.readBlob({ oid, filepath, ...options })
+  // deno-lint-ignore no-explicit-any
+  const oid = await git.resolveRef({ ref: 'git-annex', ...options } as any)
```
rwblair, line 75: Looks like resolveRef doesn't use the cache. It also expects 5 keys in the object, and only in deeper function calls do defaults for optional options get set. Can see now why Typescript is nooo having fun.

src/schema/context.test.ts
```diff
@@ -20,14 +19,14 @@ export async function makeBIDSContext(
 Deno.test('test context LoadSidecar', async (t) => {
   const context = await makeBIDSContext(dataFile)
   await t.step('sidecar overwrites correct fields', () => {
-    const { rootOverwrite, subOverwrite } = context.sidecar
+    const { rootOverwrite: _rootOverwrite, subOverwrite: _subOverwrite } = context.sidecar
```
rwblair: I don't think this line is needed, unpacking doesn't do anything special to the state of sidecar does it? Test passed without it when I tested it.
```suggestion
```

src/schema/context.test.ts
```diff
@@ -20,14 +19,14 @@ export async function makeBIDSContext(
...
     assertObjectMatch(context.sidecar, {
       rootOverwrite: 'anat',
       subOverwrite: 'anat',
     })
   })
   await t.step('sidecar adds new fields at each level', () => {
-    const { rootValue, subValue, anatValue } = context.sidecar
+    const { rootValue: _rootValue, subValue: _subValue, anatValue: _anatValue } = context.sidecar
```
rwblair: I don't think this line is needed either.
```suggestion
```

rwblair (approved):

PR #380, opened 2026-04-10
feat: Report broken symlinks

effigies:
This introduces five new error codes:

| code | severity | description|
|--|--|--|
| SYMLINK_BROKEN | error | Symbolic link target does not exist. |
| SYMLINK_CYCLE | error | Symbolic link chain contains a cycle or exceeds maximum depth |
| SYMLINK_OUT_OF_TREE | error | Symbolic link target escapes the repository root. |
| SYMLINK_IN_SUBMODULE | warning | Symbolic link target lies within an uninitialized git submodule. |
| SYMLINK_DIRECTORY_UNSUPPORTED |warning | Symbolic link to a directory is not yet supported when reading from a git tree. |

Broken symlinks and symlink cycles are a general problem on the filesystem or in a git tree. (Annex keys are not considered broken.) Out-of-tree symlinks are considered an error for git trees but not filesystem crawls. Symlinks in submodules are a problem for git trees; for filesystem crawls, this is a subset of broken symlink.

The last one is a placeholder to keep the scope of this PR a little smaller. Directories are supported for filesystem crawls, but handling them correctly in git will require some additional thought.

This fixes #374, which could have been triggering either of two independent bugs in symlink resolution.

PR #384, opened 2026-04-15
feat: Implement directory symlinks in git and verify equivalence to filesystem

effigies:
In #380, we treated symlinks to directories as broken, in order to limit the scope. This PR demonstrates how big that scope change turned out to be.

There were two goals here:

1) Have in-tree directory symlinks behave identically in a git tree as in a filesystem crawl of a checked out repository.
2) Work in a way that will facilitate adding submodule support later.

The central data structures are:

```ts
export interface TreeSource {
  readonly commitOid: string
  readonly gitOptions: GitOptions
  readonly symlinkMap: ReadonlyMap<string, string>
  readonly preferredRemote?: string
}

export type ResolveVerdict =
  | { kind: 'file-blob'; oid: string; size: number; source: TreeSource }
  | { kind: 'annex'; key: string; size: number; source: TreeSource }
  | { kind: 'tree'; treePath: string; originalDir: string; source: TreeSource }
  | {
    kind: 'submodule-boundary'
    mountPath: string
    remainder: string
    source: TreeSource
  }
  | { kind: 'unresolved'; reason: SymlinkReason }
```

A `TreeSource` is passed around to allow git operations to be run while resolving symlinks and walking directories, which are the basic actions needed to implement directory links, and this will be able to be swapped out as we eventually cross submodule boundaries. The `ResolveVerdict` enumerates the kind of things that can be at the other end of a symlink. The thing that makes all of this complicated is that you can have cases like:

```
a/b/link -> ../c/e/file
a/c -> ../d
d/e/file
```

The git tree needs to expand to three resolved blobs and their parents:

```
a/b/link
a/c/e/file
d/e/file
```

It feels a bit sprawly, so I want to see if I can be more concise. Will keep it in draft until I come have another look at it with fresh eyes.

### DIFF
--- a/src/files/gitResolver.ts
+++ b/src/files/gitResolver.ts
@@ -333,6 +333,25 @@ async function walkSubtree(
       }
       const target = source.symlinkMap.get(entryPath)
       if (target === undefined) continue
+      // Annex pointer symlinks are identified from the raw target string
+      // and do not need chain resolution. This mirrors the top-level walk
+      // in readGitTree, which short-circuits annex pointers before any
+      // resolver call.
+      const annexParsed = parseAnnexKey(target)
+      if (annexParsed !== null) {
+        result.files.push(
+          new BIDSFile(
+            outPath,
+            new AnnexedGitFileOpener(
+              annexParsed.key,
+              annexParsed.size,
+              source.gitOptions,
+              source.preferredRemote,
+            ),
+          ),
+        )
+        continue
+      }
       // Resolve nested symlinks against their ORIGINAL path (entryPath),
       // not their grafted path (outPath). Unix physical-path semantics.
       const subVerdict = await resolveSymlink(entryPath, target, source, budget, prune)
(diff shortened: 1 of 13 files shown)


src/files/gitResolver.ts
```diff
@@ -0,0 +1,394 @@
...
+      if (accumulated === '.') accumulated = ''
+      continue
+    }
+    const candidate = accumulated === '' ? seg : `${accumulated}/${seg}`
+
+    // Intermediate or terminal symlink? Consult symlinkMap first.
+    const nextTarget = source.symlinkMap.get(candidate)
+    if (nextTarget !== undefined) {
+      if (budget.remaining <= 0) {
+        return { kind: 'unresolved', reason: 'cycle' }
```
rwblair, line 174: Technically not a cycle, max depth reached. I'll leave it up to you if it needs to be a different error given how rare it is. If the max depth isn't advertised in any where other than the code I might be confused hitting the limit. But I would also have 11+ chained symlinks and probably deserve what ever difficulty the world gives me for it.

rwblair (approved):

src/files/gitResolver.ts
effigies: Right, we mostly don't try to detect cycles directly, but infer them from max depth reached. I don't want to extend this PR by trying to distinguish, but feel free to follow up with one that does, or provides a better error message.

PR #389, opened 2026-04-21
fix: Compute opaque directories from actual DatasetType

yarikoptic:
Apparently

- https://github.com/bids-standard/bids-specification/pull/2191 (attn @nikhil153 @julia-pfarr @effigies )

was merged without testing effect on the validator. That lead to the validation of my

- https://github.com/bids-standard/bids-examples/pull/451

where I moved to use such `rawbids/`, to fail!    Took a few rounds with claude to figure out the underlying cause and address it (below from claude):

`BIDSContextDataset.opaqueDirectories` was hardcoded to `rules.directories.raw`, regardless of the dataset's `DatasetType`. This had two consequences:

- Derivative datasets did not recognize derivative-only opaque directories (e.g. `template`, `cohort`).
- Study datasets did not recognize study-only opaque directories (e.g. `rawbids/` added in bids-standard/bids-specification#2191), so the walker descended into them and raised `NOT_INCLUDED` on their contents.

Switch to `rules.directories[DatasetType]` so each dataset uses the rules for its own type.  `DatasetType` is only populated after the constructor runs (the caller assigns it from the loaded `dataset_description.json` afterward), so move the computation into a helper that is invoked from both the constructor and the `dataset_description` setter, keeping the set in sync.

PS there is also a tiny doc fix I could not go by since `sourcedata/code` is misleading !

### DIFF
--- a/changelog.d/20260421_095017_yarikoptic_bf_rawbids.md
+++ b/changelog.d/20260421_095017_yarikoptic_bf_rawbids.md
@@ -1,0 +1,10 @@
+### Fixed
+
+- Compute opaque directories from the rules for the dataset's actual
+  `DatasetType` instead of always using `rules.directories.raw`.  This
+  lets study-type datasets correctly ignore the `rawbids/`
+  subdataset (added in
+  [bids-standard/bids-specification#2191](https://github.com/bids-standard/bids-specification/pull/2191))
+  and lets derivative-type datasets recognize derivative-only opaque
+  directories (e.g. `template`, `cohort`).  See
+  [#389](https://github.com/bids-standard/bids-validator/pull/389).
--- a/changelog.d/20260421_095017_yarikoptic_bf_rawbids.md
+++ b/changelog.d/20260421_095017_yarikoptic_bf_rawbids.md
@@ -3,8 +3,8 @@
 - Compute opaque directories from the rules for the dataset's actual
   `DatasetType` instead of always using `rules.directories.raw`.  This
   lets study-type datasets correctly ignore the `rawbids/`
-  subdataset (added in
-  [bids-standard/bids-specification#2191](https://github.com/bids-standard/bids-specification/pull/2191))
-  and lets derivative-type datasets recognize derivative-only opaque
-  directories (e.g. `template`, `cohort`).  See
-  [#389](https://github.com/bids-standard/bids-validator/pull/389).
+  subdataset (added in [bids-standard/bids-specification#2191][]).
+  Fixed in [#389][].
+
+[bids-standard/bids-specification#2191]: https://github.com/bids-standard/bids-specification/pull/2191
+[#389]: https://github.com/bids-standard/bids-validator/pull/389
--- a/src/schema/context.ts
+++ b/src/schema/context.ts
@@ -67,15 +67,20 @@ export class BIDSContextDataset implements Dataset {
           ?.filter((ext) => ext.endsWith('/'))
         : [],
     )
-    this.opaqueDirectories = new Set<string>(
-      args.schema
-        ? Object.values(this.schema.rules.directories.raw)
-          ?.filter((rule) => rule?.opaque && 'name' in rule)
-          ?.map((dir) => `/${dir.name}`)
-        : [],
-    )
+    this.opaqueDirectories = new Set<string>()
+    this.#refreshOpaqueDirectories()
     // @ts-expect-error Subjects type does not include null, but null is used as a sentinel for "not yet loaded"
     this.subjects = args.subjects || null
+  }
+
+  #refreshOpaqueDirectories(): void {
+    if (!this.schema?.rules?.directories) return
+    const datasetType = (this.dataset_description.DatasetType ?? 'raw') as string
+    this.opaqueDirectories = new Set<string>(
+      Object.values(this.schema.rules.directories[datasetType] ?? {})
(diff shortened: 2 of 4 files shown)


src/schema/context.ts
```diff
@@ -67,17 +67,22 @@ export class BIDSContextDataset implements Dataset {
...
     )
-    this.opaqueDirectories = new Set<string>(
-      args.schema
-        ? Object.values(this.schema.rules.directories.raw)
-          ?.filter((rule) => rule?.opaque && 'name' in rule)
-          ?.map((dir) => `/${dir.name}`)
-        : [],
-    )
+    this.opaqueDirectories = new Set<string>()
+    this.#refreshOpaqueDirectories()
```
effigies: Given that `opaqueDirectories` is being set when `this.dataset_description = args.dataset_description || {}` is called (L53), this is redundant.

src/schema/context.ts
```diff
@@ -67,17 +67,22 @@ export class BIDSContextDataset implements Dataset {
...
 
+  #refreshOpaqueDirectories(): void {
+    if (!this.schema?.rules?.directories) return
+    const datasetType = (this.dataset_description.DatasetType ?? 'raw') as string
+    this.opaqueDirectories = new Set<string>(
+      Object.values(this.schema.rules.directories[datasetType] ?? {})
+        ?.filter((rule) => rule?.opaque && 'name' in rule)
+        ?.map((dir) => `/${dir.name}`),
+    )
+  }
```
effigies: This could be inlined into `set dataset_description()`.

effigies (commented):
Good catch. Could you add a test? You should be able to do something like:

```ts
const schema = {'rules': {'directories': {
  'raw': {'a': {'opaque': true}, 'b': {'opaque': false}},
  'study': {'c': {'opaque': true}, 'd': {'opaque': false}},
}}}
const rawDataset = BIDSContextDataset({schema})
// opaqueDirectories is populated on initial load
assertEquals(rawDataset.opaqueDirectories.has('a'), true)
assertEquals(rawDataset.opaqueDirectories.has('b'), false)
assertEquals(rawDataset.opaqueDirectories.has('c'), false)
// Reset on dataset_description update
rawDataset.dataset_description = {'DatasetType': 'study'}
assertEquals(rawDataset.opaqueDirectories.has('c'), true)
assertEquals(rawDataset.opaqueDirectories.has('a'), false)
assertEquals(rawDataset.opaqueDirectories.has('d'), false)

const studyDataset = BIDSContextDataset({schema, dataset_description: {'DatasetType': 'study'}})
// Initial load respects explicit type
assertEquals(studyDataset.opaqueDirectories.has('c'), true)
assertEquals(studyDataset.opaqueDirectories.has('a'), false)
assertEquals(studyDataset.opaqueDirectories.has('d'), false)
// Reset selects correct type
studyDataset.dataset_description = {'DatasetType': 'raw'}
assertEquals(studyDatasetDataset.opaqueDirectories.has('a'), true)
assertEquals(studyDatasetataset.opaqueDirectories.has('b'), false)
assertEquals(studyDataset.opaqueDirectories.has('c'), false)
```

src/schema/context.ts
yarikoptic-gitmate: Good point — addressed in c0e6966b `refactor: Inline opaqueDirectories computation into dataset_description setter`. The setter is already called via `this.dataset_description = args.dataset_description || {}` on L53, so the extra init and call were dropped.

src/schema/context.ts
yarikoptic-gitmate: Done in c0e6966b `refactor: Inline opaqueDirectories computation into dataset_description setter`. The helper is gone; the computation now lives directly in the setter.

yarikoptic-gitmate:
Thanks for the review! Tests added in c0e6966b `refactor: Inline opaqueDirectories computation into dataset_description setter`, closely following your template. Coverage:

- default (raw) `DatasetType` populates raw opaque dirs on construction
- updating `dataset_description` switches to the new type's opaque dirs
- explicit `DatasetType: 'derivative'` on construction
- resetting to `raw` restores raw opaque dirs
- `GeneratedBy` implies `'derivative'` when `DatasetType` is missing

Also applied both inline review suggestions (removed redundant init, inlined the helper into the setter).

changelog.d/20260421_095017_yarikoptic_bf_rawbids.md
```diff
@@ -0,0 +1,10 @@
+### Fixed
+
+- Compute opaque directories from the rules for the dataset's actual
+  `DatasetType` instead of always using `rules.directories.raw`.  This
+  lets study-type datasets correctly ignore the `rawbids/`
+  subdataset (added in
+  [bids-standard/bids-specification#2191](https://github.com/bids-standard/bids-specification/pull/2191))
+  and lets derivative-type datasets recognize derivative-only opaque
+  directories (e.g. `template`, `cohort`).  See
+  [#389](https://github.com/bids-standard/bids-validator/pull/389).
```
effigies: Template and cohort are not opaque. Using reference links instead of inline for readability.

```suggestion
  subdataset (added in [bids-standard/bids-specification#2191][]).
  Fixed in [#389][].

[bids-standard/bids-specification#2191]: https://github.com/bids-standard/bids-specification/pull/2191
[#389]: https://github.com/bids-standard/bids-validator/pull/389
```

effigies (approved):

julia-pfarr (approved):

PR #390, opened 2026-04-21
feat: Recurse into nested BIDS datasets beyond derivatives/, and normalize invalid DatasetType

yarikoptic:
Co-Authored-By: Claude Code 2.1.116 / Claude Opus 4.7 (1M context) <noreply@anthropic.com>

## Summary

Two orthogonal-but-related fixes around nested BIDS datasets and the
`DatasetType` field:

1. **Recurse into nested BIDS datasets under `rawbids/` and `sourcedata/`,
   not just `derivatives/`**. The previous recursion logic scanned only
   `derivatives/<name>/dataset_description.json`. It silently ignored
   two other shapes that appear in real datasets:
   - Single nested: the container directory itself is a BIDS dataset      (e.g. `rawbids/dataset_description.json`, which was added as a      study-level container in  [bids-standard/bids-specification#2191](https://github.com/bids-standard/bids-specification/pull/2191),     and the historical `sourcedata/raw/dataset_description.json` form).
   - Multiple nested: sub-directories of the container are each BIDS
     datasets (e.g. `derivatives/fmriprep/`,     `sourcedata/ds00003/dataset_description.json` as used by OpenNeuro
     study-style layouts).

   With this change, `-r` now discovers and validates all of these.
   Bugs planted inside a `rawbids/` or `sourcedata/ds00003/` subdataset
   are caught; previously they went silently undetected.

2. **Normalize an invalid `DatasetType` for internal rule lookups**.
   The setter previously defaulted only when `DatasetType` was falsy; a
   truthy-but-out-of-enum value (e.g. `"Raw"` — capitalization typo in
   `bids-examples/proj-nipoppy/rawbids/dataset_description.json`) made
   `rules.directories[DatasetType]` resolve to `undefined`, cascading
   into spurious `NOT_INCLUDED` errors on legitimate `sub-*/ses-*/`
   subjects on top of the legitimate `JSON_SCHEMA_VALIDATION_ERROR`.

   Fall back to the spec default for any out-of-enum value. Valid enum
   is pulled from `schema.objects.metadata.DatasetType.enum` (no
   hardcoded list). Normalization is done on a shallow copy so the
   memoized JSON is untouched — JSON schema validation still flags the
   enum violation exactly once.

## User-visible behavior changes

- `-r` now recurses into nested BIDS datasets found under any of
  `derivatives/`, `rawbids/`, or `sourcedata/`, at either the container's
  own level or one level down.
- Text output label `"Derivative:"` is now `"Nested dataset:"`.
- The `derivativesSummary` field on `ValidationResult` is preserved for
  API stability but now covers all nested BIDS datasets (not only
  derivatives). Documented in the `validate` JSDoc.
- Datasets with an out-of-enum `DatasetType` still get the
  `JSON_SCHEMA_VALIDATION_ERROR` they already got, but no longer
  cascade into spurious `NOT_INCLUDED` on legitimate subjects.

## Not in this PR (flagged for follow-up)

- Compliance checks currently unenforced: a `raw` (or `derivative`) root
  containing a nested BIDS dataset under `rawbids/` or `sourcedata/`,
  and a non-`study` root containing `rawbids/` at all. Likely best paired
  with an upstream schema proposal to declare nested-BIDS capability on
  directory rules (e.g. `nested_bids: single | multiple | none`), so
  the validator stops hardcoding the container set.
  (@yarikoptic has even missed that and needs to get back to that PR to see if we discussed this or not to be also part of 'derivatives' dataset type!)
- **Pre-existing**: `filememoize` (`src/utils/memoize.ts`) keys by
  `(file.parent.path, file.path)`, which collides across rerooted trees.
  When multiple nested BIDS datasets are validated in parallel under
  `-r`, whichever `loadJSON` runs first wins and others reuse its
  content. Affects multi-derivative recursion too. Tracked separately in #391 

## Test plan

- [x] `deno test --allow-env --allow-read --allow-write --allow-run --allow-net src/` — 336/336 pass.
- [x] Synthetic dataset reproducing the nested-rawbids + invalid-`DatasetType` case:
  - `-r` surfaces the planted bug inside `rawbids/sub-002/bogus_file.txt`.
  - `-r` surfaces the planted bug inside `sourcedata/ds00003/sub-A/stray.dat`.
  - `JSON_SCHEMA_VALIDATION_ERROR` still fires for `DatasetType: "Raw"`.
  - No spurious `NOT_INCLUDED` on legitimate `sub-001`.
- [x] Existing derivatives recursion behavior unchanged (covered by
  `src/tests/local/derivatives.test.ts`).


src/schema/context.ts
```diff
@@ -83,12 +83,31 @@ export class BIDSContextDataset implements Dataset {
...
+    if (!copy.DatasetType) {
+      copy.DatasetType = copy.GeneratedBy ? 'derivative' : 'raw'
+    } else {
+      // For a truthy value, cross-check against the schema's enum when one is
+      // available.  A mismatch (e.g. "Raw" capitalization typo) would make
+      // `rules.directories[DatasetType]` resolve to undefined and cascade into
+      // spurious NOT_INCLUDED errors; fall back to the default as above.  If
+      // the schema isn't loaded (test construction without a schema) we can't
+      // check — leave the value untouched.
+      // @ts-expect-error `metadata` is not declared on SchemaObjects; access is dynamic
```
yarikoptic: let me know if I should reduce commenting here

src/validators/bids.ts
```diff
@@ -126,23 +130,30 @@ export async function validate(
     }
   }
 
-  const bidsDerivatives: Promise<FileTree>[] = []
-  const nonstdDerivatives: FileTree[] = []
+  // Directories that may contain nested BIDS datasets — either the directory
+  // itself is a BIDS dataset (e.g. `rawbids/dataset_description.json`) or its
+  // immediate subdirectories are (e.g. `derivatives/fmriprep/`,
+  // `sourcedata/ds00003/`).
+  const nestingContainers = ['derivatives', 'rawbids', 'sourcedata']
```
yarikoptic: for this to be schema driven we first need to adjust schema AFAIK since we do not have indicators there on where to expect subdatasets (opaque or not)

src/schema/context.ts
effigies: It's definitely a lot. I don't think we need the branching logic, though, `validTypes.includes(undefined)` will also return `False`.

```ts
// shallow copy to avoid polluting memoize cache
value = {...value}
const validTypes = this.schema?.objects?.metadata?.DatasetType?.enum as Array
// Unknown types (including capitalizations) fall back to defaults
// Per spec: 'the default value is `"raw"`'.
// Before DatasetType was defined, `GeneratedBy` was used as a derivatives indicator
if (validTypes && !validTypes.includes(value.DatasetType as string)) {
  value.DatasetType = value.GeneratedBy ? 'derivative' : 'raw'
}
this.#dataset_description = value
```

I do wonder if we should drop the `GeneratedBy` behavior, since the spec does not say to do this, and `GeneratedBy` is now valid in raw datasets.

src/validators/bids.ts
effigies: Agreed. Let's think about how to do it in the schema.

src/schema/context.ts
yarikoptic: yeah, I thought about that too! Moreover [GeneratedBy](https://bids-specification.readthedocs.io/en/stable/glossary.html#objects.metadata.GeneratedBy) is imho very applicable to "raw" datasets as they are typically generated by some tool + human(s) and thus we did provide it for "raw" (made it explicit!)  BIDS dataset and I think frankly it should even be RECOMMENDED for "raw" (not study) : https://github.com/OpenNeuroDatasets/ds005256/blob/main/dataset_description.json#L37

yarikoptic: so let's kick it out in a separate PR?

src/validators/bids.ts
yarikoptic: in parallel to providing hardcoded solution here or instead? (thus overall probably just moving this PR in Draft)

effigies: In parallel. We don't need to hold up validator features on the schema. Something working today is better than waiting on a schema release.

src/validators/bids.ts
```diff
@@ -126,23 +130,30 @@ export async function validate(
...
-        bidsDerivatives.push(subtree(deriv))
-      } else {
-        nonstdDerivatives.push(deriv)
+    if (dir.get('dataset_description.json')) {
+      bidsNestedDatasets.push(subtree(dir))
+    } else {
+      for (const sub of dir.directories) {
+        if (sub.get('dataset_description.json')) {
+          bidsNestedDatasets.push(subtree(sub))
+        }
```
effigies, line 166: We shouldn't lose these. We don't do it now, but we should be including files in there for total dataset file count/size.

```suggestion
        } else {
          nonstdDerivatives.push(deriv)
        }
```

src/schema/context.ts
effigies: Sure, can be a separate PR.

changelog.d/20260421_095017_yarikoptic_bf_recurse.md
```diff
@@ -0,0 +1,22 @@
...
+
+- Nested BIDS datasets discovered under `derivatives/`, `rawbids/`, or
+  `sourcedata/` are now recursed into with `-r`, whether the container
+  directory itself is a BIDS dataset (e.g. `rawbids/dataset_description.json`)
+  or it holds one or more BIDS datasets as immediate subdirectories
+  (e.g. `derivatives/fmriprep/dataset_description.json`,
+  `sourcedata/ds00003/dataset_description.json`). The text output
+  section previously labeled "Derivative:" is now "Nested dataset:".
+  The `derivativesSummary` field on `ValidationResult` is preserved
+  for API stability but now covers all nested BIDS datasets.  See
```
yarikoptic: this made me pause -- should we separate out into `sourcedatasetsSummary` (would include `rawbids/`) from `derivativesSummary`?  it would add a field but overall be "compatible" if anyone uses that, right @effigies ?

src/schema/context.ts
yarikoptic: since we want this one to land and not wait -- just did right here.

src/validators/bids.ts
yarikoptic: done

changelog.d/20260421_095017_yarikoptic_bf_recurse.md
yarikoptic: decided to give claude a try for this, pushing... moving to draft - will need to test etc

yarikoptic: I moved this back from draft -- should be ready for review/test and potentially merge!

- https://github.com/bids-standard/bids-validator/issues/416

is related and a solution would be a reasonable follow up  to this one

src/schema/context.ts
yarikoptic: could be marked resolved @effigies ?

PR #395, opened 2026-04-29
feat(web): Allow users to pass custom schema to web validator

effigies:
It turns out using custom schemas on Windows is painful (cc @melanieganz for links to the known problems). This is a quick hack to allow loading custom schemas in the web validator.

The error reporting is not well styled, but this gets the job done.


effigies:
Post-merge reviews welcome. Will create an issue to improve styling.

effigies:
@melanieganz This is now live on https://bids-standard.github.io/bids-validator/dev/

PR #399, opened 2026-05-04
Don't generate sidecarkey required errors for derivative datasets fromm non derivative related sidecar rules.

rwblair:
This skips the question of `schema.rules.json` entries. There are currently no rules there that explicitly apply to derivatives. Are the rules in there metadata?

One issue is we are relying on derivatives sub directory in `schema.rules.sidecars` If rules ever get reorganized it will cause a regression.

### DIFF
--- a/src/schema/applyRules.ts
+++ b/src/schema/applyRules.ts
@@ -216,8 +216,7 @@ function evalJsonCheck(
        */
       if (
         context.dataset.dataset_description.DatasetType === 'derivative' &&
-        schemaPath.startsWith('rules.sidecars') &&
-        !schemaPath.startsWith('rules.sidecars.derivatives')
+        !rule.selectors.includes('dataset.dataset_description.DatasetType == "derivative"')
       ) {
         continue
       }
--- a/src/schema/applyRules.ts
+++ b/src/schema/applyRules.ts
@@ -216,7 +216,7 @@ function evalJsonCheck(
        */
       if (
         context.dataset.dataset_description.DatasetType === 'derivative' &&
-        !rule.selectors.includes('dataset.dataset_description.DatasetType == "derivative"')
+        !rule.selectors?.includes('dataset.dataset_description.DatasetType == "derivative"')
       ) {
         continue
       }
--- a/src/schema/applyRules.ts
+++ b/src/schema/applyRules.ts
@@ -205,28 +205,43 @@ function evalJsonCheck(
 
     if (value === undefined) {
       const severity = getFieldSeverity(requirement, context)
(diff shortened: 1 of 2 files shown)


src/schema/applyRules.ts
```diff
@@ -205,28 +205,43 @@ function evalJsonCheck(
...
+      /*
+       * quoth derivatives introduction in the specification:
+       *
+       * "Unless specified otherwise, individual sidecar JSON files and all metadata fields within are OPTIONAL."
+       */
+      if (
+        context.dataset.dataset_description.DatasetType === 'derivative' &&
+        schemaPath.startsWith('rules.sidecars') &&
+        !schemaPath.startsWith('rules.sidecars.derivatives')
+      ) {
```
effigies, line 220: Do we want to hard-code structure like this? We could skip if `dataset.dataset_description.DatasetType` doesn't appear in any selectors. It seems safe to assume that, if it does, the selectors will select the right things.

rwblair: I had thought about this, I agree that it is more generalizable. Two concerns that popped into my head were:

1. CPU cycles, though when has this ever stopped us before?
2. Hard coding the selector could run into the same issues some day. `dataset.dataset_description.DatasetType == "derivative"` is the vast majority of selectors on datasettype but the expression language is flexible enough that a rule could be written with `intersects` or some other function.

I'm happy to rewrite and filter based on selectors if you aren't worried about the above.

effigies: 1. Fair. I think pruning the schema might be the more efficient move, but that will require more careful thought.
2. Yes, but the result would still depend on the value of `dataset.dataset_description.DatasetType`, which means we can leave the logic to the rule author. Or do you see a case where that might not hold?

I think we should do it selector-based, either here during the walk or in an initial prune. If pruning is something to work on soon, then we could keep the structural approach for now, since it works.

src/schema/applyRules.ts
```diff
@@ -205,28 +205,42 @@ function evalJsonCheck(
...
+      }
+
+      /*
+       * quoth derivatives introduction in the specification:
+       *
+       * "Unless specified otherwise, individual sidecar JSON files and all metadata fields within are OPTIONAL."
+       */
+      if (
+        context.dataset.dataset_description.DatasetType === 'derivative' &&
+        !rule.selectors.includes('dataset.dataset_description.DatasetType == "derivative"')
```
effigies:
```suggestion
        !rule.selectors?.includes('dataset.dataset_description.DatasetType == "derivative"')
```

PR #408, opened 2026-05-18
Minor test corrections to accommodate windows

VisLab:
## Summary

Fixes five tests that fail on Windows due to CRLF line endings and unconditional `chmod` calls.

### 1. CRLF line endings in `bom-utf8.json`

The BOM-stripping test in `deno.test.ts` compared `.text()` output against a string joined with the platform `EOL` constant from `@std/fs`. On Windows `EOL` is `\r\n`, but `bom-utf8.json` is stored in git with LF-only line endings. With `core.autocrlf=true`, git converts the file to CRLF on checkout, causing the assertion to fail.

### 2. `chmod` called on Windows in `finally` blocks

Three `finally` blocks called `chmod -R +w` unconditionally as part of temp-directory cleanup. `chmod` does not exist on Windows, so these calls threw `NotFound`, prevented `Deno.remove()` from running, and caused the tests to fail.

## Changes

| File | Change |
|---|---|
| `.gitattributes` | Mark `src/tests/bom-utf8.json` as `text eol=lf` so git enforces LF for all contributors regardless of `core.autocrlf` |
| `src/files/deno.test.ts` | Replace `join(EOL)` with `join('\n')` and remove the now-unused `EOL` import — the fixture has a known fixed encoding, platform-adaptive EOL is wrong here |
| `src/files/utils.test.ts` | Guard the `chmod` call in `withRepo` with `if (!isWindows)` |
| `src/files/git.test.ts` | Guard the `chmod` calls in the `finally` blocks of the ds000001 integration test and the bare-repository test with `if (!isWindows)` |

## Tests

Previously failing on Windows:

- `Deno implementation of BIDSFile … strips BOM characters when reading UTF-8 via .text()`
- `readGitTree: error on nonexistent ref`
- `readGitTree: abbreviated SHA resolution`
- `readGitTree: pruning directories`
- `readGitTree: bare repository`

All five now pass. The `chmod` calls still run on Linux/macOS (the guard is Windows-only), so behaviour there is unchanged.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,1 +1,2 @@
 bids-validator/src/.git-meta.json  export-subst
+src/tests/bom-utf8.json  text eol=lf
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -55,10 +55,13 @@ for their contributions to this release.
 
 ## Fixed
 
-- Test suite now passes on Windows: `src/tests/bom-utf8.json` is pinned to LF
-  line endings via `.gitattributes` (previously checked out as CRLF, breaking the
-  BOM-stripping assertion), and `chmod -R +w` cleanup calls in
-  `src/files/git.test.ts` and `src/files/utils.test.ts` are skipped on Windows
+- Test suite now passes on Windows: `src/tests/bom-utf8.json`
+  is pinned to LF line endings via `.gitattributes` 
+  (previously checked out as CRLF, breaking the
+  BOM-stripping assertion).
+  
+- `chmod -R +w` cleanup calls in`src/files/git.test.ts` and 
+  `src/files/utils.test.ts` are skipped on Windows
   where `chmod` is not available.
 
 - Use absolute path to `.js` within Dockerfile ENTRYPOINT,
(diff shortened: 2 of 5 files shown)


effigies:
https://github.com/bids-standard/bids-validator/blob/main/.github/CONTRIBUTING.md#changelog-entries

> #### Changelog entries
> 
> We are using the [scriv] changelog management tool to help keep our changelog organized.
> 
> Most pull requests will deserve changelog entries, and some may deserve more than one.
> Scriv allows changelog entries to be made as part of the pull request, which gives
> the contributor and the reviewer a chance to consider the impact of the pull request.
> 
> To add an entry, run:
> 
> ```
> scriv create --edit
> ```
> 
> (If you're wondering how to install `scriv`, we recommend using [`uvx scriv`][uvx].)
> 
> Your editor will open a Markdown file with commented sections.
> Find the appropriate section(s) for your contribution and write a 1-3 line description.

[scriv]: https://scriv.readthedocs.io/en/latest/
[uvx]: https://docs.astral.sh/uv/guides/tools/

PR #409, opened 2026-05-18
Fixed React double-invocation cleanup type error

VisLab:


VisLab:
@effigies  Could you explain the scriv error more fully?  I did run the "uvx scriv create --edit".  However that left a weird directory with the fragments. I did the `scriv collect` to merge into the CHANGELOG.  If that special `scriv_check.sh` script or the `validation.yml` will actually delete these entries and do the `scriv collect` -- I couldn't see it.  Please rename the `validation.yml` workflow and put more complete instructions. If the validation error is something else, let me know. I would like to get this merged so I can update the `hed-validator` that is being used to 4.2.0. This change removes the HED browser based validator from the `hed-validator` and corrects a few minor issues. Thx

effigies:
@VisLab The idea is just to check the fragments in (e.g., dbc1f854216acd7fd530283a57caadf7601428b4), and then we collect them at release (e.g., ebd55fb5a7a87205c7d5819467ba68566fb10d9e). By using fragments, there is no chance of merge conflicts when multiple PRs are proposing fixes/features/etc.

web/src/App.tsx
```diff
@@ -28,9 +28,9 @@ function Files({ issues }) {
   }));
   return (
     <ul>
-      {[...unique.values()].map(({ location, issueMessage }) => {
+      {[...unique.entries()].map(([key, { location, issueMessage }]) => {
         return (
-          <li key={location}>
+          <li key={key}>
```
effigies, line 33: Could you screenshot/paste what issue you were seeing here and the impact of the change?

VisLab: The bug was revealed during a copilot code review.  Here is a detailed explanation (sorry I'm asked Claude for this). In my next commit I'm adding a test for it called `src/files-dedup.test.ts`

# Bug Fix: Duplicate React Keys in the Web Validator's File Issue List

## Summary

A bug in the BIDS web validator caused some per-file issue messages to be silently
dropped from the UI when multiple distinct validation messages were reported for the
same file path. The fix was a two-line change to how React list-item keys are generated
in the `Files` component.

---

## The Component

The `Files` component renders the list of files affected by a particular validation
issue. Each entry optionally includes an `issueMessage` — a short per-occurrence detail
(e.g. which field is missing). Before rendering, exact duplicates are collapsed using a
`Map` keyed by the composite string `${location}${issueMessage ?? ""}`.

```tsx
function Files({ issues }) {
  const unique = new Map(issues.map(({ location, issueMessage }) => {
    return [`${location}${issueMessage ?? ""}`, { location, issueMessage }];
  }));
  return (
    <ul>
      {[...unique.entries()].map(([key, { location, issueMessage }]) => (
        <li key={key}>...</li>
      ))}
    </ul>
  );
}
```

---

## The Bug

In the original code the `.map()` iterated over `unique.values()` and used `location`
alone as the React `key`:

```tsx
// BEFORE (buggy)
{[...unique.values()].map(({ location, issueMessage }) => (
  <li key={location}>...</li>
))}
```

**Why this is wrong:** React uses the `key` prop to track which DOM elements correspond
to which data items across renders. If two items share the same key, React treats them as
the *same* element and only renders one of them.

Consider a dataset where one file violates two different rules:

| `location` | `issueMessage` |
|---|---|
| `/sub-01/anat/sub-01_T1w.nii.gz` | `missing field 'foo'` |
| `/sub-01/anat/sub-01_T1w.nii.gz` | `missing field 'bar'` |

The `Map` correctly retains both entries (their composite keys differ). But when rendered
with `key={location}`, both `<li>` elements would share the key
`/sub-01/anat/sub-01_T1w.nii.gz`. React would:

1. Log a console warning: *"Encountered two children with the same key."*
2. Silently discard one of the two entries, so the user never sees the second message.
3. Potentially update the wrong DOM node during re-renders.

---

## The Fix

Iterate over `unique.entries()` instead, and use the *map key* (the composite string)
as the React key:

```tsx
// AFTER (fixed)
{[...unique.entries()].map(([key, { location, issueMessage }]) => (
  <li key={key}>...</li>
))}
```

The map key is the same composite string used to deduplicate in the first place, so
it is **guaranteed to be unique** across all rendered items.

---

## Was This Caught by a Test?

No. The bug was identified through code review, not an automated test failure. The web
project had no component-level rendering tests at the time.

VisLab:
Persistence pays off!

PR #413, opened 2026-06-03
fix: Prune dotfiles rather than ignore, make `--prune` inject default ignores into prune list

### LINKED ISSUE #412: Talks to git too much?!  takes forever on slimmed down clone
We were trying to validate against a PR schema a trimmed down fork of https://datasets.datalad.org/?dir=/dbic/QA where we left just one subject/session.  But it seems to take forever (I will report when it finishes) and strace shows lots of interaction with git:

```
reprostim@typhon:~/branches/QA-reprostim$ datalad run -- duct bids-validator-deno --schema https://bids-specification--2367.org.readthedocs.build/en/2367/schema.json  -o derivatives/bids-validator.txt --debug DEBUG --ignoreNiftiHeaders .
[INFO   ] == Command start (output follows) =====
2026-06-02T14:31:33-0400 [INFO    ] con-duct: duct 0.17.0 is executing 'bids-validator-deno --schema https://bids-specification--2367.org.readthedocs.build/en/2367/schema.json -o derivatives/bids-validator.txt --debug DEBUG --ignoreNiftiHeaders .'...
2026-06-02T14:31:33-0400 [INFO    ] con-duct: Log files will be written to .duct/logs/2026.06.02T14.31.33-329389_
....
```

`top` shows 200-300% CPU by bids-validator (2.4.1) and

```
(reprostim) reprostim@typhon:~/branches/QA-reprostim$ strace -f -p 329392 2>&1 | grep '\.git' | head
[pid 329392] readlink("/home/reprostim/branches/QA-reprostim/.git", 0x7ffe0ddb5a80, 1023) = -1 EINVAL (Invalid argument)
[pid 329392] readlink("/home/reprostim/branches/QA-reprostim/.git/objects", 0x7ffe0ddb5a80, 1023) = -1 EINVAL (Invalid argument)
[pid 329392] readlink("/home/reprostim/branches/QA-reprostim/.git/objects/pack", 0x7ffe0ddb5a80, 1023) = -1 EINVAL (Invalid argument)
[pid 329392] readlink("/home/reprostim/branches/QA-reprostim/.git/objects/pack/pack-f9e5d018d993267cb05737418a73360dca7e40f5.pack", 0x7ffe0ddb5a80, 1023) = -1 EINVAL (Invalid argument)
[pid 329392] openat(AT_FDCWD, "/home/reprostim/branches/QA-reprostim/.git/objects/pack/pack-f9e5d018d993267cb05737418a73360dca7e40f5.pack", O_RDONLY|O_NOFOLLOW|O_CLOEXEC) = 12
[pid 329392] readlink("/home/reprostim/branches/QA-reprostim/.git",  <unfinished ...>
[pid 329392] readlink("/home/reprostim/branches/QA-reprostim/.git/objects",  <unfinished ...>
[pid 329392] readlink("/home/reprostim/branches/QA-reprostim/.git/objects/90",  <unfinished ...>
[pid 329392] readlink("/home/reprostim/branches/QA-reprostim/.git/objects/90/4d7742b418eb73474438138a00a11144332818",  <unfinished ...>
[pid 329392] readlink("/home/reprostim/branches/QA-reprostim/.git",  <unfinished ...>
```

what is it trying to 'discover' ?

effigies:
Initially narrowly scoped to prune `.git`, this PR now includes the following changes:

* Dotfiles and directories are now pruned during walk, rather than ignored after the walk. This saves descent into `.git/` or other potentially deep hidden directories (`.venv/`, for example).
* `--prune` now injects the default ignore list into the prune list, which is a more principled way to decide what gets pruned.
* To preserve the original intended behavior of `--prune`, `derivatives/` is now ignored.
* To avoid pruning `.bidsignore` and causing `--prune` to un-ignore everything, `.bidsignore` is now ignored through a carve-out in the `BIDSFile.ignored` property. As a validator feature and not a BIDS feature, this seems reasonable to me.
* The file tree builders now all construct a default prune list for consistency.

Open issues (probably for separate PRs):

* Generating the default ignore list from schema. There's a bit of a chicken and egg problem where if the ignore list comes from a newer schema than is declared in `dataset_description.json`, we're not respecting that, but that's already the case. We already don't load the schema based on `dataset_description.json`, so as long as we load it from `--schema`, we're still improving.
* The file count skips ignored files...  With significant confounds like `.git` now being pruned instead of ignored, I think it may be time to revisit this and count ignored files for a more accurate description of dataset size.

Closes #412.

### DIFF
--- a/src/main.ts
+++ b/src/main.ts
@@ -24,9 +24,13 @@ export async function main(): Promise<ValidationResult> {
   setupLogging(options.debug)
 
   const absolutePath = resolve(options.datasetPath)
-  const prune = options.prune
-    ? new FileIgnoreRules(['derivatives', 'sourcedata', 'code'], false)
-    : undefined
+  const prune = new FileIgnoreRules([], 'prune')
+  if (options.prune) {
+    // Opaque BIDS directories are ignored by default but still included to allow lookups,
+    // including size estimates and symlink resolution in git trees.
+    // Pruning is an extreme measure to reduce memory usage and IO ops and may have unintended consequences.
+    prune.addDefaults('ignore')
+  }
   const tree = options.gitRef
     ? await readGitTree(absolutePath, options.gitRef, prune, options.preferredRemote)
     : await readFileTree(absolutePath, prune, options.preferredRemote)
--- a/changelog.d/20260602_211743_markiewicz_prune_git.md
+++ b/changelog.d/20260602_211743_markiewicz_prune_git.md
@@ -1,0 +1,11 @@
+### Changed
+
+- The `--prune` option removes opaque BIDS directories (`sourcedata/`, `derivatives/`, etc.)
+  from the file tree, reducing processing time and memory usage.
+  This option is incompatible with recursive validation and may cause unexpected
+  behavior if pruned data are referenced by name or via symbolic links.
+
+### Fixed
+
+- `.dotfiles` and directories (such as `.git/`) are now pruned from the file tree,
+  avoiding unnecessary IO operations and memory consumption.
--- a/src/api/filetree/mod.ts
+++ b/src/api/filetree/mod.ts
@@ -15,3 +15,4 @@
 export { FileTree } from '../../types/filetree.ts'
 export { filesToTree, loadBidsIgnore, subtree } from '../../files/filetree.ts'
 export { FileIgnoreRules, readBidsIgnore } from '../../files/ignore.ts'
+export type { IgnoreGroup, LogOptions } from '../../files/ignore.ts'
--- a/src/files/browser.ts
+++ b/src/files/browser.ts
@@ -41,7 +41,7 @@ export function fileListToTree(
   return loadBidsIgnore(
     filesToTree(
       files.map((f) => new BIDSFileBrowser(f, ignore, root))
-        .filter((f) => !prune.test(f.path)),
+        .filter((f) => !prune.test(f.path, { log: true, prefix: 'Pruned' })),
       ignore,
     ),
     ignore,
--- a/src/files/browser.ts
+++ b/src/files/browser.ts
@@ -31,11 +31,19 @@ export class BIDSFileBrowser extends BIDSFile {
  * @param files - Files selected via the browser file picker.
  * @returns The root `FileTree` with `.bidsignore` rules applied.
  */
-export function fileListToTree(files: File[]): Promise<FileTree> {
+export function fileListToTree(
+  files: File[],
+  prune?: FileIgnoreRules,
+): Promise<FileTree> {
+  prune ??= new FileIgnoreRules([], 'prune')
   const ignore = new FileIgnoreRules([])
   const root = new FileTree('/', '/', undefined)
   return loadBidsIgnore(
-    filesToTree(files.map((f) => new BIDSFileBrowser(f, ignore, root)), ignore),
+    filesToTree(
+      files.map((f) => new BIDSFileBrowser(f, ignore, root))
+        .filter((f) => !prune.test(f.path)),
+      ignore,
+    ),
     ignore,
   )
 }
--- a/src/files/deno.ts
+++ b/src/files/deno.ts
@@ -71,7 +71,7 @@ async function _readFileTree({
 
   for await (const dirEntry of Deno.readDir(join(rootPath, relativePath))) {
     const thisPath = posix.join(relativePath, dirEntry.name)
-    if (prune.test(thisPath)) {
+    if (prune.test(thisPath, { log: true, prefix: 'Pruned' })) {
       continue
     }
 
--- a/src/files/deno.ts
+++ b/src/files/deno.ts
@@ -155,7 +155,7 @@ export async function readFileTree(
   prune?: FileIgnoreRules,
   preferredRemote?: string,
 ): Promise<FileTree> {
-  prune ??= new FileIgnoreRules([], false)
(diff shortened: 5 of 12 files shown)


effigies:
@yarikoptic Care to give this branch a try?

src/main.ts
```diff
@@ -24,9 +24,10 @@ export async function main(): Promise<ValidationResult> {
   setupLogging(options.debug)
 
   const absolutePath = resolve(options.datasetPath)
-  const prune = options.prune
-    ? new FileIgnoreRules(['derivatives', 'sourcedata', 'code'], false)
-    : undefined
+  const prune = new FileIgnoreRules(['.git**'], false)
```
yarikoptic: isn't bids-validator ignoring all `.dotdirs`? why then only `.git` to ignore and not everything which starts with `.`?

src/main.ts
```diff
@@ -24,9 +24,10 @@ export async function main(): Promise<ValidationResult> {
...
+  if (options.prune) {
+    prune.add(['derivatives', 'sourcedata', 'code'])
```
yarikoptic: although unrelated to this PR, but I guess should be also 


```suggestion
    prune.add(['code', 'derivatives', 'docs', 'logs', 'rawbids', 'sourcedata'])
```

src/main.ts
effigies: Pruned directories are omitted from the size, ignored directories are not. This is why we do not prune ignored directories by default. I think `.git` is categorically different and including its size double-counts and it should have always been pruned.

src/main.ts
effigies: cc @rwblair who wrote the pruning originally. Any reason not to extend this? rawbids should not be in the `main` branch until 1.11.2 or 1.12.0 are released. It could go into `dev`, since it's been merged. I agree with #416 however that this should be schema-driven, which is why I kept this change narrow to address the problem.

effigies: Actually, as I think about this, I think we chose derivatives/sourcedata/code because they are the most likely to have submodules that could extend the file tree walk significantly when checked out. I'm not sure that logs or docs have the same issues. I think it might instead make sense to allow `--prune` to take an argument that accepts a list of opaque directories, and default to derivatives/sourcedata/code (+rawbids, eventually).

src/main.ts
yarikoptic: oh, shady area... Frankly -- I do not see `.git` much different in this case from `.tox` (I have GBs caches and do get one for "bids" datasets under `code/`), `.cache/` or potentially many other `.dotdirs`.  So to that end, I still do not see why to discriminate `.git` from other .dotdirs.  Overall, I do not think that bids-validator's job is to somehow provide "trusty stats" -- it is to validate, and BIDS ignores dotdirs https://bids-specification.readthedocs.io/en/stable/common-principles.html#dotfiles :

> Any file or directory starting with a . present in a BIDS dataset is considered hidden and not subject to BIDS validation.

Hence I would exclude all `.dotdirs`!  Might worth clarification in docs at most that hidden folders are excluded.  Having said that there could be an option (like `--include-paths REGEX` for whitelisting) if some tool (like `openneuro`) relies on those size reporting to avoid double scanning etc.

And for the git-annex'ed repos, I wonder how size then is computed - as extracted from git-annex keys or dereferenced to `.git/annex` symlinks? (may be that is why somehow .git was included?)

effigies: I can get behind pruning dotdirs.

src/main.ts
yarikoptic: > I think we chose derivatives/sourcedata/code because they are the most likely to have submodules that could extend the file tree walk significantly when checked out. I'm not sure that logs or docs have the same issues.

FWIW, next Yarik's "wisdom" (in addition to *duplication is evil*): *no assumptions is likely the best assumption!* ;) 

There could be some api refs under `docs/` and heavy submodules (like repronim/containers and alike) under `code/` for self-containment/reproducibility.  I would just stay consistent with the schema as much as possible and avoid assumptions.

effigies:
Well, this turned into a rabbit hole. Will try again later.

effigies:
PR description updated. I cast the review net a bit wider, since this is a fairly significant change.

src/main.ts
```diff
@@ -24,9 +24,13 @@ export async function main(): Promise<ValidationResult> {
...
 
   const absolutePath = resolve(options.datasetPath)
-  const prune = options.prune
-    ? new FileIgnoreRules(['derivatives', 'sourcedata', 'code'], false)
-    : undefined
+  const prune = new FileIgnoreRules([], 'prune')
+  if (options.prune) {
+    // Opaque BIDS directories are ignored by default but still included to allow lookups,
+    // including size estimates and symlink resolution in git trees.
+    // Pruning is an extreme measure to reduce memory usage and IO ops and may have unintended consequences.
```
rwblair, line 31: This may need to be added to the help text for the flag.

rwblair (approved):
I'm waffling on how transparent we should be with what exactly is pruned when the `--prune` option isn't used or is false. The defaults are sensible so I don't think we'll surprise any one, but my lack of imagination has tied my shoelaces together before.

Should the contents of `ignoreDefaults` be more visible in help text or documentation?

Do you think a `logger.info` readout of pruned files would be useful to anyone?

The nomeclature/semantics of `--prune` feels strange sometimes, but something like `--extended-prune` feels worse, so I'm fine with keeping it for now.

PR #420, opened 2026-06-12
fix: correctly parse headerless _motion.tsv files

### LINKED ISSUE #394: TSV_COLUMN_HEADER_DUPLICATE reported for valid headerless *_motion.tsv files
Hi, I am validating a BIDS Motion dataset and the validator reports
`TSV_COLUMN_HEADER_DUPLICATE` for `*_motion.tsv` files.

The files are headerless. According to the BIDS Motion specification, motion
files must not have a header row, and column order is defined by the associated
`*_channels.tsv` file:

The error appears when the first data row contains repeated values such as
`0`, `0.0`, `-1.0`, or `NaN`. It looks like the validator may be applying a
generic TSV header check to a file whose first row is data, not headers.

Example start of first data row:
```tsv
NaN	0.0	-1.0	-1.0	-1.0	0	0	0.0	0.0
```

Validator command:

```bash
deno run -ERWN jsr:@bids/validator /path/to/output
```

The error message:
```text
[ERROR] TSV_COLUMN_HEADER_DUPLICATE Two elements in the first row of a TSV are the same.
```
Is this expected behavior for BIDS Motion files, or should *_motion.tsv be exempt from this header-duplicate check?

Yehuda-Bergstein:
Fixes #394                                                                                  
                                                                                                
### Description                                                                             
This PR fixes a bug where the Deno validator incorrectly parsed the first row of data in    
  `_motion.tsv` files as column headers, resulting in false-positive                            
  `CUSTOM_COLUMN_WITHOUT_DESCRIPTION` errors.                                                   
                                                                                                
According to BEP029, `_motion.tsv` files are uncompressed and explicitly headerless. However,
the standard `_loadTSV` function was previously hardcoded to always consume the first row as  
headers.                                                                                      
                                                                                                
### Changes Made                                                                            
1. **`src/files/tsv.ts`**: Introduced a `headerless` boolean parameter (default `false`) to 
2. `_loadTSV` and its memoized wrappers. When `headerless` is true, the parser safely bypasses   
duplicate header checks, auto-generates dummy columns to preserve table structure, and retains
the first row of numerical data.                                                              
3. **`src/schema/context.ts`**: Inside `BIDSContext.loadColumns()`, if the file suffix is   
`motion`, it now explicitly calls the TSV parser with `headerless: true`. After structural    
validation, it dynamically clears the parsed `columns` map for motion files so that           
`evalColumns` ignores them (mirroring the legacy validator's behavior).                       
4. **`src/schema/associations.ts` & Tests**: Updated existing data loaders and test calls to
explicitly pass `false` to maintain standard behavior for all other BIDS TSVs.                
                                                                                                
### Testing                                                                                 
- Ran validation locally against real-world VR tracking datasets containing `_motion.tsv`   
files; verified the custom column errors disappeared.                                         
- Successfully ran the full test suite (`deno task test` against the `bids-examples` submodule) with 349 passing tests and 0 failures.


src/schema/context.ts
```diff
@@ -267,6 +268,9 @@ export class BIDSContext implements Context {
           logger.debug(error)
           return new Map<string, string[]>() as ColumnsMap
         }) as Record<string, string[]>
+      if (headerless) {
+        this.columns = new ColumnsMap() as Record<string, string[]>
+      }
```
effigies: Here you undo the load. That's not going to work.

effigies (changes requested):
Thanks for this! I think the cleaner patch would actually be to the `loadTSVGZ` immediately above where we make the `.pipeThrough(new DecompressionStream('gzip'))` call optional.

I would suggest we give it a more general name, e.g., `loadHeaderlessTSV`, with a compression parameter. You'll also need to extract the headers from the associated `channels.tsv` file, which will involve adding a `name` field to:

https://github.com/bids-standard/bids-validator/blob/f0cf0fc63f6d1c78da42d6d04e262de86a4384fb/src/schema/associations.ts#L93-L104

You can then pass in `this.associations.channels.name` as the headers.

src/schema/context.ts
Yehuda-Bergstein: Reworked it like you suggested, `loadTSVGZ` is now `loadHeaderlessTSV` with a compression flag and the motion headers come from the channels.tsv `name` column via the association (18846aa).

One thing I ran into is that `Channels` from `@bids/schema` has no `name` field so it wouldn't compile. I added a local type alias (`ChannelsWithName`) to get around it but not sure that's the right way. Should `name` be added to the schema context model instead?

effigies: Both are good. A PR against the spec to add the field would be great, but this is a good way to get quick support in the validator.

effigies (approved):
All looks good. I updated `context.ts` to deduplicate the headerless case of `loadColumns()`, and reworded the changelog to more affirmatively describe the fix (*We now do X* instead of *We no longer do Y*). Merging on passing tests.

effigies:
Merged! Please add yourself to https://github.com/bids-standard/bids-validator/blob/main/CITATION.cff if you would like to be credited as a contributor. You can skip the changelog for that one.

PR #425, opened 2026-06-30
feat(wheel): Publish single universal wheel, depending on deno package

### LINKED ISSUE #423: Lighter weight wheels
https://pypi.org/project/deno/ now exists. We might be able to get away with a single wheel that depends on `deno` and uses much less disk space (including our PyPI limit). I think the way to do this would be to run `deno cache` using the build directory as the cache, then set the script with something like:

```python
#!python
import os
import os.path as op
import sys

from deno import find_deno_bin


def cli():
    exe = find_deno_bin()
    win = sys.platform.startswith('win')
    env = {'DENO_DIR': op.join(op.dirname(__file__), 'DENO_DIR')}
    permissions = [
        # Access filesystem for reading BIDS datasets
        "--allow-read",
        # Accept environment variables
        "--allow-env",
        # Access network for fetching remote schemas
        "--allow-net",
        # Allow writing results to file
        "--allow-write",
        # Run git to get version info
        "--allow-run=git",
    ]
    if win:
        # Terminal detection code in supports-hyperlinks calls osRelease
        permissions.append("--allow-sys=osRelease")

    argv = ['deno', *permissions, f'jsr:@bids/validator' *sys.argv[1:]]

    if sys.platform.startswith('win'):
        exec_subproc(exe, argv)
    else:
        os.execv(exe, sys.argv)


def exec_subproc(executable, argv):
    import subprocess

    try:
        subprocess.run(
            argv,
            executable=executable,
            shell=False,
            check=True,
        )
        # try flush here to trigger a BrokenPipeError
        # within the try-except block so we can handle it
        # (happens if the calling process closed stdout
        # already)
        sys.stdout.flush()
    except BrokenPipeError:
        # setting it to None prevents Python from trying to
        # flush again
        sys.stdout = None
    except subprocess.CalledProcessError as e:
        sys.exit(e.returncode)
```

Much of this taken from https://github.com/psychoinformatics-de/git-annex-wheel.

effigies:
This drops the wheel size from 50MB / platform / release to <5MB / release. The basic idea is that it uses `deno bundle` to create a `bids-validator.js` file. Then we inject an `__init__.py` that finds the deno binary and runs the bundle with appropriate permissions.

Closes #423. Builds on #424 to verify that it still runs on all tested platforms.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -7,12 +7,12 @@ authors = [
 dependencies = ["deno>=2.5.0"]
 readme = "README.md"
 requires-python = ">=3.10"
-license = {text = "MIT"}
+license = "MIT"
+license-files = ["LICENSE"]
 classifiers = [
     "Development Status :: 5 - Production/Stable",
     "Intended Audience :: Science/Research",
     "Topic :: Scientific/Engineering :: Bio-Informatics",
-    "License :: OSI Approved :: MIT License",
     "Programming Language :: JavaScript",
 ]
 keywords = ["BIDS", "BIDS validator"]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -4,8 +4,9 @@ description = "Typescript implementation of the BIDS validator"
 authors = [
     {name = "bids-standard developers"},
 ]
-dependencies = []
+dependencies = ["deno>=2.5.0"]
 readme = "README.md"
+requires-python = ">=3.10"
 license = {text = "MIT"}
 classifiers = [
     "Development Status :: 5 - Production/Stable",
@@ -22,6 +23,9 @@ Documentation = "https://bids-validator.readthedocs.io/"
 "Source code" = "https://github.com/bids-standard/bids-validator"
 Issues = "https://github.com/bids-standard/bids-validator/issues"
 
+[project.scripts]
+bids-validator-deno = "bids_validator_deno:cli"
+
 [dependency-groups]
 doc = [
     "sphinx",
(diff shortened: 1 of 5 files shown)


effigies:
Hey, that wasn't too bad. IMO the main point for discussion is what people think about generating `__init__.py` from `pdm_build.py`. I think it's cleaner to have all of the Python wrapping in a single file, rather than now also constructing a Python source directory, but others may disagree.

@candleindark WDYT about this overall?

Before merging, we need to include an appropriate license declaration, since the generated code is based off of [`git_annex/__init__.py`](https://github.com/psychoinformatics-de/git-annex-wheel/blob/main/src/git_annex/__init__.py) which is [AGPL 3](https://github.com/psychoinformatics-de/git-annex-wheel/blob/3330a8bb46206281809a9501ba5881186480304a/COPYRIGHT#L852-L855). @mih I'm thinking that including a notice in LICENSE and in the generated code template (so it gets included in the generated `__init__.py`) would satisfy the license, but I want to check with you. The overall project is MIT-licensed. The status of `pdm_build.py` itself is unclear to me, since it quotes the copyleft material rather than executes it.

effigies:
Pushed licensing updates.

<details><summary>The generated file looks like:</summary>

```python
# This module was generated at build time by pdm_build.py.
#
# SPDX-License-Identifier: AGPL-3.0-or-later
#
# This module was adapted from git-annex-wheel/src/git_annex/__init__.py @ commit 9a059be:
#
# https://github.com/psychoinformatics-de/git-annex-wheel/blob/9a059be12db8b90cb1ecf71e498d3a2135366b6f/src/git_annex/__init__.py
# https://hub.datalad.org/git-annex/git-annex-wheel/src/commit/9a059be12db8b90cb1ecf71e498d3a2135366b6f/src/git_annex/__init__.py
#
# Changes:
#   * Adapted cli() to find Deno bundled in a separate package and exec the local bundle.
"""Find the deno runtime and exec the bundled BIDS validator."""
import os
import sys

def cli():
    from deno import find_deno_bin

    deno = os.fsdecode(find_deno_bin())
    windows = sys.platform.startswith('win')
    bundle = os.path.join(os.path.dirname(__file__), 'bids-validator.js')

    permissions = ['--allow-read', '--allow-env', '--allow-net', '--allow-write', '--allow-run=git']
    if windows:
        permissions.extend(['--allow-sys=osRelease'])

    argv = [deno, *permissions, bundle, *sys.argv[1:]]

    if windows:
        exec_subproc(deno, argv)
    else:
        os.execv(deno, argv)


def exec_subproc(executable, argv):
    import subprocess

    try:
        subprocess.run(
            argv,
            executable=executable,
            shell=False,
            check=True,
        )
        # try flush here to trigger a BrokenPipeError within the try-except block
        # (happens if the calling process closed stdout already)
        sys.stdout.flush()
    except BrokenPipeError:
        # setting to None prevents Python from trying to flush again
        sys.stdout = None
    except subprocess.CalledProcessError as e:
        sys.exit(e.returncode)
```

</details>

mih:
@effigies given that the two of us are the only people that contributed to the code you based this work on: If you agree (and I do), we dual-license it under AGPL-3 and MIT. My original motivation was homogeneity with git-annex. Nothing speaks against also establishing homogeneity in this context.

Given then simplicity of the code, I doubt it is necessary to make a big statement somewhere. From my POV this issue comment should be good enough to move forward, and only acknowledge the origin in the LICENSE statement.

pyproject.toml
```diff
@@ -32,41 +36,12 @@ doc = [
 ]
 
 [build-system]
-requires = ["pdm-backend"]
+requires = ["pdm-backend", "deno>=2.5"]
```
candleindark:
```suggestion
requires = ["pdm-backend", "deno>=2.5.0"]
```

To be cosmetically consistent with the dependencies specification without changing the meaning.

Issue #434, opened 2026-07-10
Warn on presence of deprecated metadata fields


psadil:
"deprecated" is a valid requirement_level, according to the metaschema: https://github.com/bids-standard/bids-specification/blob/4235ae7a2a523b657f502b7b0c852093585d7614/src/metaschema.json#L554-L558

> ```
>   "requirement_level": {
>       "$comment": "Requirement levels that may apply to terms",
>       "type": "string",
>       "enum": ["required", "recommended", "optional", "deprecated"]
>     }
> ```

So, with no "deprecated" option, the "severity" of a deprecated field is set to "undefined". The old block was skipped when the guard evaluated to false (as "undefined" would cause it to), meaning no warning was emitted

https://github.com/bids-standard/bids-validator/blob/024f6879f112f93ff7c5d26a38d9a376966606e2/src/schema/applyRules.ts#L208-L230

[It] seems the mapping from requirement_level to `severity` should be schema-driven, rather than hard-coded from a list of strings. But reverting that guard will also address the issue.

*Note: The above was edited by @effigies to focus on the remaining issue, which is that deprecated fields fall through the logical cracks. The motivating bug was fixed in #436.*

---

<details><summary>Original post</summary>

This is limited to the pre-release and not the latest stable (silly me for working with alpha). 

Current alpha release ([AcquisitionDuration is deprecated](https://github.com/bids-standard/bids-schema/blob/1fbd48df925d60686a191838e1affd5cb8df1b7f/versions/latest/schema/rules/sidecars/func.yaml#L66-L70))

```bash
❯ deno run -A lib/bids-validator/src/bids-validator.ts --version
bids-validator 3.0.0-alpha.4-2-g6eaae4e9

❯ deno run -A lib/bids-validator/src/bids-validator.ts tests/data/bids-examples/ds003 --ignoreNiftiHeaders --json | jq | rg -B1  AcquisitionDuration 
        "code": "SIDECAR_KEY_RECOMMENDED",
        "subCode": "AcquisitionDuration",
--
        "code": "SIDECAR_KEY_RECOMMENDED",
        "subCode": "AcquisitionDuration",
--
        "code": "SIDECAR_KEY_RECOMMENDED",
        "subCode": "AcquisitionDuration",
--
        "code": "SIDECAR_KEY_RECOMMENDED",
        "subCode": "AcquisitionDuration",
--
        "code": "SIDECAR_KEY_RECOMMENDED",
[...]

PR #436, opened 2026-07-10
fix: Fully invert logical condition

effigies:
This narrowly addresses the issue in #434, however it does not address @psadil's larger point that deprecated fields should be addressed.

b6635d3 inverted an if block to continue, to allow for more intricate logic afterwards. The full expression was not inverted, leading to cases where deprecated fields were behaving like recommended fields, because deprecated mapped onto undefined.

Handling deprecations is its own feature that needs consideration, so for now just complete the logic inversion.


effigies:
Forgoing changelog, since this bug was introduced in an unreleased PR.

PR #440, opened 2026-07-17
fix: Check hashDirMixed paths in AnnexedGitFileOpener

nellh:
Support hashDirMixed in AnnexedGitFileOpener for local files. This matches the git-annex behavior of checking hashDirMixed first before falling back to hashDirLower, but it is more stat calls than just branching on whether or not it's a bare repository. It's possible to have a long lived git-annex repo that uses both so this is the most compatible method.


effigies (approved):
LGTM. Thanks!

PR #442, opened 2026-08-21
Fix: Correct syntax checking for HED column

happy5214:
The code checking for the existence of a `HED` column incorrectly used the object-style `in` syntax instead of the `Map`-style `has` method. This was making the check non-functional, resulting in datasets containing a `HED` column but no sidecar HED data to not be validated by hed-validator. This also appears to be causing the error mentioned in https://github.com/hed-standard/hed-javascript/issues/836. This commit fixes the check.

### DIFF
--- a/changelog.d/20260827_150743_rosswilsonblair_validate_hed_with_hed_column.md
+++ b/changelog.d/20260827_150743_rosswilsonblair_validate_hed_with_hed_column.md
@@ -1,0 +1,47 @@
+<!--
+A new scriv changelog fragment.
+
+Uncomment the section that is right (remove the HTML comment wrapper).
+For top level release notes, leave all the headers commented out.
+-->
+
+<!--
+### Added
+
+- A bullet item for the Added category.
+
+-->
+<!--
+### Changed
+
+- A bullet item for the Changed category.
+
+-->
+### Fixed
+
+- Issue with existence of HED column in context being improperly verified.
+
+<!--
+### Deprecated
+
+- A bullet item for the Deprecated category.
+
+-->
+<!--
+### Removed
+
+- A bullet item for the Removed category.
+
+-->
+<!--
+### Security
+
+- A bullet item for the Security category.
+
+-->
+<!--
+### Infrastructure
+
+- A bullet item for the Infrastructure category.
+
+-->
--- a/pdm_build.py
+++ b/pdm_build.py
@@ -46,7 +46,7 @@ def pdm_build_initialize(context):
             deno,
             "bundle",
             "--frozen",
-            "--sourcemap",
+            "--sourcemap=linked",
             "-o",
             str(bundle),
             "src/bids-validator.ts",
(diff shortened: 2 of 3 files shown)


happy5214:
OK, that didn't pass type-checking. Am I supposed to subscript the object or something? (Is there a reason for the type cast?)

ttm:
Reporter of hed-standard/hed-javascript#836 here. Answering your type-check
question, since the one-line change is right about the bug and only wrong about
which interface to reach for.

**Why the cast exists.** `BIDSContext` declares the field as a plain object, not
as the Map:

```ts
// src/schema/context.ts
columns: Record<string, string[]>                              // line 135
this.columns = new ColumnsMap() as Record<string, string[]>    // line 169
```

So statically `context.columns` is a `Record<string, string[]>`, and
`context.columns.has` resolves through the index signature to `string[]`. Hence
"Type 'string[]' has no call signatures". At runtime the object really is a
`ColumnsMap` proxy and `.has()` does work, which is why the behaviour was right
and the compiler still refused.

Reproduced from the declaration alone, TypeScript 5.7:

```ts
declare const columns: Record<string, string[]>
'HED' in columns                       // ok
columns.has('HED')                     // TS2349, same error as CI
Object.keys(columns).includes('HED')   // ok
columns['HED'] !== undefined           // ok
```

**Minimal fix.** Either of the last two type-checks and matches what
`src/schema/tables.ts` already does with this field (`Object.keys(context.columns)`
at lines 227 and 326, `context.columns[name] as string[]` at 233 and 378):

```ts
(context.columns['HED'] !== undefined || sidecarHasHed(context.sidecar))
```

**The root cause, if you want it.** The reason `'HED' in context.columns` never
fired is the proxy's `has` trap in `src/types/columns.ts`:

```ts
has: function (target: ColumnsMap, prop: string) {
  return Reflect.has(target, prop)
}
```

`Reflect.has` inspects properties of the underlying `Map` object, but column data
lives in Map *entries*, so `in` misses every column. Running the proxy exactly as
written, with columns `onset`, `duration`, `HED`:

```
'HED' in columns          ->  false     (the column exists)
'has' in columns          ->  true      (not a column)
'forEach' in columns      ->  true      (not a column)
'constructor' in columns  ->  true      (not a column)
```

So `in` is wrong in both directions on this type. Since the point of the proxy is
to let consumers treat columns as a `Record`, and `in` is part of that interface,
the trap is arguably the real defect:

```ts
has: function (target: ColumnsMap, prop: string) {
  return target.has(prop)
}
```

That would fix `hed.ts` without touching `hed.ts`, and it agrees with `ownKeys`,
which already reports only column names. The tradeoff is that `'size' in columns`
and friends become false, so it is a wider change than you may want inside this
PR.

Happy either way. If you take the call-site fix, hed-standard/hed-javascript#836
has six minimal probe datasets that pin all of this, and I will open them as
regression fixtures against this branch if they would be useful.

Issue #411, opened 2026-05-28
bug: Crash with Uncaught TypeError (NaN) in formatSummary when dataset contains symlinks


rcruces:
### **Description**
A regression was introduced between `v2.0.11` and `v2.4.1` causing the validator to crash with an uncaught `TypeError: Expected a finite number, got number: NaN`. 

This seems to happen specifically when evaluating datasets that utilize **symlinks**. 

### **Context / Dataset Setup**
We are currently preparing a dataset for release and utilizing symlinks to map data across different subject IDs (`sub-*`) to avoid duplicating heavy imaging files on disk. 

* **Works perfectly in:** `bids-validator v2.0.11` (Successfully resolves symlinks, reads the 7.53 GB dataset, and returns a clean validation summary).
* **Crashes in:** `bids-validator v2.4.1` (Fails during the `formatSummary` stage while calculating file/byte metrics).

---

### **Error Log (v2.4.1)**
When executing via Deno or within a Singularity container mounting the dataset, the execution panics at the summary output step:

```text
singularity exec --containall -B "${bids_dir}":"${bids_dir}" "${dcm2bids_img}" deno run --allow-write -ERN jsr:@bids/validator ${bids_dir} --ignoreWarnings --outfile ${bids_dir}/bids_validator_output.txt

╭ Warning
│
│  Ignored build scripts for packages:
│  npm:core-js-pure@3.49.0
│
│  Lifecycle scripts are only supported when using a `node_modules` directory.
│  Enable it in your deno config file:
│  "nodeModulesDir": "auto"
╰─
error: Uncaught (in promise) TypeError: Expected a finite number, got number: NaN
    throw new TypeError(`Expected a finite number, got ${typeof num}: ${num}`);
          ^
    at format ([https://jsr.io/@std/fmt/1.0.10/bytes.ts:123:11](https://jsr.io/@std/fmt/1.0.10/bytes.ts:123:11))
    at formatSummary ([https://jsr.io/@bids/validator/2.4.1/src/utils/output.ts:199:51](https://jsr.io/@bids/validator/2.4.1/src/utils/output.ts:199:51))
    at consoleFormat ([https://jsr.io/@bids/validator/2.4.1/src/utils/output.ts:36:15](https://jsr.io/@bids/validator/2.4.1/src/utils/output.ts:36:15))
[...]

PR #443, opened 2026-08-28
Fixes validator crash on NaN summary size for symlinked datasets

brainy-joseph:
Prevent crash described in Issue #411 

When a dataset contains symlinks, `summary.size` can be NaN, and `prettyBytes` (from @std/fmt/bytes) throws "TypeError: Expected a finite number, got number: NaN", crashing the validator at its summary step.

This guards the value with `Number.isFinite`, falling back to 0 so the summary always renders. Display only change and has no validation logic touched. Thank you! :D

### DIFF
--- a/changelog.d/fix_nan_summary.md
+++ b/changelog.d/fix_nan_summary.md
@@ -1,0 +1,2 @@
+### Fixes
+* Fixed validator crash on NaN summary size for symlinked datasets.
--- a/src/summary/summary.ts
+++ b/src/summary/summary.ts
@@ -114,7 +114,10 @@ export class Summary {
     }
 
     this.totalFiles++
-    this.size += await context.file.size
+      const fileSize = await context.file.size
+      if (Number.isFinite(fileSize)){
+        this.size += fileSize
+      }
 
     if ('sub' in context.entities) {
       this.subjects.add(context.entities.sub)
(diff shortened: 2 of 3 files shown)


brainy-joseph:
@pre-commit-ci please autofix!

effigies:
I think we probably want to fix this by avoiding adding `nan` to the summary size in the first place, not detecting it at the end. Then we at least still get the size of the resolvable files.

I'm guessing the `nan` happens on broken symlinks, not just any symlink?

brainy-joseph:
> I think we probably want to fix this by avoiding adding `nan` to the summary size in the first place, not detecting it at the end. Then we at least still get the size of the resolvable files.
> 
> I'm guessing the `nan` happens on broken symlinks, not just any symlink?

Hi @effigies ,  I have now updated the PR to fix this at the source!

I reverted the display only update in ⁠output.ts⁠ and instead added a ⁠Number.isFinite⁠ check in ⁠src/summary/summary.ts⁠ where the size is accumulated. This way, valid file sizes are still counted but NaN values from broken symlinks are skipped.

Let me know if this looks good!

brainy-joseph:
pre-commit.ci autofix

effigies (approved):
Could you add a small change log entry? See the contributing guide for how to use scriv.

brainy-joseph:
> Could you add a small change log entry? See the contributing guide for how to use scriv.

@effigies  Read the Contribution Guide and added the changelog entry manually via GitHub as I am currently out of my city. I hope that was what you were looking for. Let me know if any changes are needed! :)

